PR #4, opened 2017-11-10
Fix host IP formatting and improve local IP API

isaachier:


isaachier:
Just changed the names, not the actual functions. Not sure if it needs a new test but I'll get back to you soon.

isaachier:
Decided it was a good idea to add one anyway. Posted in most recent commit.

PR #5, opened 2017-11-13
Crossdock

isaachier:
Not necessarily ready yet. Not sure how to test locally. @yurishkuro @black-adder, am I on the right track with `Dockerfile` and `docker-compose.yml`? This is all new territory for me. Thanks!


isaachier:
Actually, going to use `rules.mk` as reference for testing. Will get back to you soon.

isaachier:
Let me finish this and repost it then.

PR #7, opened 2017-11-14
Backport crossdock fixes

isaachier:


src/jaegertracing/samplers/RemotelyControlledSampler.cpp
```diff
@@ -42,14 +43,21 @@ class HTTPSamplingManager : public sampling_manager::thrift::SamplingManagerIf {
...
+            _serverAddr = socket.connect(serverURL);
+        } catch (...) {
+            utils::ErrorUtil::logError(_logger, "cannot connect to socket");
+        }
     }
 
     void getSamplingStrategy(SamplingStrategyResponse& result,
                              const std::string& serviceName) override
     {
+        if (_serverAddr == net::IPAddress()) {
```
black-adder: hmm, not following. _serverAddr is set to net::IPAddress() if the client couldn't reach agent?

.gitignore
```diff
@@ -1,5 +1,7 @@
-# Build directory
+# Build files
 /build
+/crossdock/crossdock
+/crossdock/jaeger-docker-compose.yml
```
black-adder: this a temporary thing? don't think we should be ignoring this completely

isaachier: crossdock/crossdock is the executable binary itself. Actually changed this in my recent crossdock development. The `jaeger-docker-compose.yml` is a downloaded file to launch the other jaeger clients.

src/jaegertracing/samplers/RemotelyControlledSampler.cpp
isaachier: Yes. `net::IPAddress()` is the default value for an empty IP and the address is only set upon successful connection. The reason for this is to avoid contacting the server if we couldn't connect on construction. I guess the real solution is not to start the looping thread at all, but that is another class that ends up calling this periodically.

isaachier:
I'm closing this in favor of real crossdock pull request

PR #9, opened 2017-11-15
Overflow bug fix

isaachier:
@rnburn does this make sense? I wish C++ had `ullround`.


rnburn:
That's definitely not the right fix. `std::llround` will round to the nearest `int64_t`, and `std::numeric_limits<int64_t>::max()` is roughly half `std::numeric_limits<uint64_t>::max()`, so you'll be effectively sampling at 50%.

isaachier:
Right that's what I feared. Well I see hhvm dealt with a similar issue and used a ternary statement to resolve it. See https://github.com/facebook/hhvm/issues/5932. Let me see if I can use that.

rnburn:
It looks like that's for signed conversion.

I would cast your sampling date to a `long double` which can perfectly represent 64-bit integers (provided it's using an 80-bit implementation). Even if you add a conditional to handle the overflow, it's not going to be a very precise way of approximating the sampling rate.

The only problem with `long double`'s is that they're required to be implemented on all platforms (e.g. [win32](https://stackoverflow.com/a/7120740/4447365)).

You might be best off just using a less clever way of sampling with `std::uniform_real_dist`

isaachier:
Then let me post this on the Go client because that is where I "stole" this idea from.

PR #18, opened 2017-11-22
Set default-constructed timestamp to current time in StartSpan

### LINKED ISSUE #14: Traces not showing up in jaeger.
I'm running this small c++ program (matching the go quickstart [example](http://opentracing.io/documentation/pages/quick-start#go)).

```cpp
#include <jaegertracing/Tracer.h>
#include <opentracing/tracer.h>
#include <chrono>
#include <thread>

int main() {
  auto config = jaegertracing::Config(
      false,
      jaegertracing::samplers::Config(jaegertracing::kSamplerTypeConst, 1),
      jaegertracing::reporters::Config(
          jaegertracing::reporters::Config::kDefaultQueueSize,
          std::chrono::seconds(1), true));
  auto tracer = jaegertracing::Tracer::make("wherearemyspans?", config);
  auto span = tracer->StartSpan("abc");
  std::this_thread::sleep_for(std::chrono::milliseconds{10});
  span->Finish();
  std::this_thread::sleep_for(std::chrono::seconds{5});
  tracer->Close();
  return 0;
}
```

If I start the Jaeger's docker server
```
docker run -d -e COLLECTOR_ZIPKIN_HTTP_PORT=9411 -p5775:5775/udp -p6831:6831/udp -p6832:6832/udp \
  -p5778:5778 -p16686:16686 -p14268:14268 -p9411:9411 jaegertracing/all-in-one:latest
```
it shows the service name "wherearemyspans?" (so the server received something); but if I try to find traces, I don't see anything.

Running the go example, the spans show up all the time.

Any idea what could be going on?

isaachier:
Fixes #14

### DIFF
--- a/src/jaegertracing/Span.cpp
+++ b/src/jaegertracing/Span.cpp
@@ -80,15 +80,19 @@ void Span::SetBaggageItem(opentracing::string_view restrictedKey,
 void Span::FinishWithOptions(
     const opentracing::FinishSpanOptions& finishSpanOptions) noexcept
 {
+    const auto finishTimeSteady =
+        (finishSpanOptions.finish_steady_timestamp ==
+          SteadyClock::time_point())
+         ? SteadyClock::now() : finishSpanOptions.finish_steady_timestamp;
     std::shared_ptr<const Tracer> tracer;
     {
+
         std::lock_guard<std::mutex> lock(_mutex);
         if (isFinished()) {
             // Already finished, so return immediately.
             return;
         }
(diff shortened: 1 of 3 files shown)


rnburn:
If one of the system/steady timestamps is specified but not the other, shouldn't this try to convert the provided timestamp instead of using now?

Also, I think this same sort of logic is missing for finish as well.

isaachier:
Good point. Let me add both of those suggestions.

PR #25, opened 2017-12-03
CMake build fixes

isaachier:


isaachier:
Since this is a build system fix, going to accept it myself.

PR #29, opened 2017-12-04
Redirect upload to jaegertracing cache

isaachier:


yurishkuro:
Is this going to commit changes to that new repo on every build? Which other projects are using this approach?

isaachier:
Not every time but every time libraries are added, a new platform is added in Travis, or the compiler mode changes. The details of the process are described [here](https://docs.hunter.sh/en/latest/faq/why-binaries-from-server-not-used.html). I took a lot of the code from [cpp-ethereum](https://github.com/ethereum/cpp-ethereum) only because it is one of the few projects I could find that leverages this feature.

isaachier:
Still making sure this works so please don't accept yet.

isaachier:
OK tested and ready to go once you approve.

yurishkuro:
I can see that https://github.com/jaegertracing/cpp-client-hunter-cache now contains two directories, is that expected? Is this number going to continue growing?

yurishkuro:
I guess I do not understand the process. Are you caching the source code of dependencies? Why can't we simply pull them from github?

isaachier:
Imagine Maven was a build tool for Java without any hosted/centralized JAR repository. Each project would either have to rebuild the JARs of its dependencies from source every time or host their own JAR repository. The same idea applies to Hunter with C/C++ libraries. It is not such a significant time difference for our project now (5 minutes with cache vs. 10 minutes without it), so if it bothers you we can remove the cache repo entirely.

yurishkuro:
Maven Central stores compiled dependencies. Is that what you're storing in the cache repo?  Why not use some other services like Quay or BinTray?

Also, Node apps often download and compile a lot of dependencies. Travis supports cache dir settings for those, couldn't we use that for C++?  How does Envoy project do that?

yurishkuro:
btw I just created https://bintray.com/jaegertracing

isaachier:
Cool. I'll take a look at Envoy and other hosting services.

isaachier:
OK now using cache directories (see https://docs.travis-ci.com/user/caching/). Thanks for the tip. The current changes should be ready to land.

Issue #30, opened 2017-12-04
Segfault in localIP function


isaachier:
```
Segmentation fault {
  libpthread-2.19.so 10330 __restore_rt
  libjaegertracingd.so.0.0.8 258bfa IPAddress.cpp:40 jaegertracing::net::IPAddress::localIP(int)::{lambda(ifaddrs const*)#1}::operator()(ifaddrs const*) const
  libjaegertracingd.so.0.0.8 25974d functional:2058 std::_Function_handler<bool (ifaddrs const*), jaegertracing::net::IPAddress::localIP(int)::{lambda(ifaddrs const*)#1}>::_M_invoke(std::_Any_data const&, ifaddrs const*)
  libjaegertracingd.so.0.0.8 259bc9 functional:2472 std::function<bool (ifaddrs const*)>::operator()(ifaddrs const*) const
  libjaegertracingd.so.0.0.8 258d04 IPAddress.cpp:50 jaegertracing::net::IPAddress::localIP(std::function<bool (ifaddrs const*)>)
  libjaegertracingd.so.0.0.8 258c51 IPAddress.cpp:41 jaegertracing::net::IPAddress::localIP(int)
  libjaegerutils.so 28813 jaegertracing::Tracer::Tracer(std::string const&, std::shared_ptr<jaegertracing::samplers::Sampler> const&, std::shared_ptr<jaegertracing::reporters::Reporter> const&, std::shared_ptr<jaegertracing::logging::Logger> const&, std::shared_ptr<jaegertracing::metrics::Metrics> const&, int)
  libjaegerutils.so 296a8 jaegertracing::Tracer::make(std::string const&, jaegertracing::Config const&, std::shared_ptr<jaegertracing::logging::Logger> const&, jaegertracing::metrics::StatsFactory&, int)
  libjaegerutils.so 1e4f1 jaegertracing::utils::make_jaeger_tracer_for_prod(std::string const&)
  libjaegerutils.so 1e8e1 jaegertracing::utils::make_jaeger_tracer(std::string const&)
  map_api_load_test 406de3 main
  libc-2.19.so 21f45 libc-start.c:321 __libc_start_main
  map_api_load_test 407dd1 _start
}
Segmentation fault (core dumped)
```

The above occurred when connected via VPN. I believe the issue is here: `ifAddr->ifa_addr->sa_family`, and I'm assuming `ifa_addr` is null.

PR #31, opened 2017-12-04
Fix bug in localIP and revert change in TR1 tuple definition

isaachier:
Signed-off-by: Isaac Hier <ihier@uber.com>


isaachier:
Fixes #30.

isaachier:
Thanks @black-adder

Issue #36, opened 2018-01-15
Error "fatal error: yaml-cpp/yaml.h2" building library. JAEGERTRACING_WITH_YAML_CPP=OFF is ignored?


carlosjgp:
Cloning the repository and building the client fails because of a missing dependency.
```
tempDir="$(mktemp -d)" \
&& chmod 777 "$tempDir" \
&& cd "$tempDir" \
&& git clone https://github.com/jaegertracing/cpp-client.git jaeger-cpp-client \
&& cd jaeger-cpp-client \
&& mkdir .build && cd .build \
&& cmake -DCMAKE_BUILD_TYPE=Release \
           -DBUILD_TESTING=OFF \
           -DJAEGERTRACING_WITH_YAML_CPP=OFF ..
```

```
In file included from /tmp/tmp.hHLAaO/jaeger-cpp-client/src/jaegertracing/Config.h:21:0,
                 from /tmp/tmp.hHLAaO/jaeger-cpp-client/src/jaegertracing/Config.cpp:17:
/tmp/tmp.hHLAaO/jaeger-cpp-client/src/jaegertracing/baggage/RestrictionsConfig.h:24:27: fatal error: yaml-cpp/yaml.h: No such file or directory
 #include <yaml-cpp/yaml.h>
                           ^
```
Full output
https://gist.github.com/carlosjgp/d58b744cc699e0e19ba8718ec359e7d1

I have tried to clone and build yaml-cpp before with the same output :(

```
tempDir="$(mktemp -d)" \
&& chmod 777 "$tempDir" \
&& cd "$tempDir" \
&& git clone https://github.com/jbeder/yaml-cpp.git \
&& cd yaml-cpp \
&& mkdir .build && cd .build \
&& cmake -DBUILD_SHARED_LIBS=ON .. \
&& cd "$tempDir" \
&& git clone https://github.com/jaegertracing/cpp-client.git jaeger-cpp-client \
&& cd jaeger-cpp-client \
&& mkdir .build && cd .build \
&& cmake -DCMAKE_BUILD_TYPE=Release \
           -DBUILD_TESTING=OFF \
           -DJAEGERTRACING_WITH_YAML_CPP=OFF ..
```

carlosjgp:
It does work if you install the yaml-cpp library and you don't use the `JAEGERTRACING_WITH_YAML_CPP` flag 
```
tempDir="$(mktemp -d)" \
&& chmod 777 "$tempDir" \
[...]

PR #39, opened 2018-01-24
Fix yaml-cpp issues (#36)

isaachier:
Signed-off-by: Isaac Hier <ihier@uber.com>


isaachier:
I used a tool to optimize my `#include`s, but it totally undermined the case when certain libraries weren't installed, so had to fix here.

Issue #45, opened 2018-01-28
Thrift IDL not in main repository, no build support


ringerc:
In the jaeger cpp-client source dir:

```
$ find . -name \*.thrift
$
```

and a `git grep emitZipkinBatch` finds only generated output from Thrift + references in `UDPClient.h`, `UDPClientTest.cpp`, and `MockAgent.h`.

Where are the Thrift sources to re-generate what's in `thrift-gen` ?

It's necessary for #38 to work since the local Thrift may not be the same version as the Thrift compiler used to check in the sources. Plus it's not great to carry generated sources in git when you can avoid it.

isaachier:
OK so the Thrift schema files are shared among many language bindings. Hence, they are actually in a submodule `idl`. I'm assuming `git grep` is skipping submodules. You can regenerate using the Makefile in that directory. I agree I am not a fan of the use of version 0.9.2, but this has more to do with Uber's internal use than anything else.

ringerc:
@isaachier Aah, that's `git`'s lovely default of silently ignoring submodules biting me. Thanks for the pointer. I'll look. Guess I'll need to glue that into the CMake build.

ringerc:
On 29 January 2018 at 02:47, isaachier <notifications@github.com> wrote:

> I don't think you need to plug in to build at all (I didn't for a reason).
> Just regenerate a move to your thrift-gen directory.
>
>
> That's fine for an ad-hoc local build. But for packaging and deployment it
should be done the same way every time, in the CMakeFile. Ideally the
generated sources shouldn't be in git at all, just the IDL to produce them,
but that's not so practical when the IDL is in another module.

It's also not user friendly to have to manually fetch some IDL, generate
some sources, move them across, etc when you do a build. Especially if
undocumented. The more of these sorts of corner cases there are the harder
it is to get someone to use a tool, contribute to it, etc.

Consider a typical workflow like this:

- Fetch commit from git, with any submodules
- "make dist" a tarball and gpg sign it, annotating it with the commit id
or tag
- (offline) generate srpm with .spec file and sources
- (offline) build the srpm on target platforms, using local repositories of
signed RPMs for dependencies
- (offline) sign the rpms
- Push resulting .rpm and debuginfo into repository for testing
- Run smoketests in CI
- Push resulting rpms to public repository for use by other
builds/tools/users

"make dist" needs to know how to generate the IDL. For users to be able to
compile the software the way they'd usually expect to, so does "make".

[...]

PR #46, opened 2018-01-28
Support building with local dependencies (no Hunter)

ringerc:
jaeger-cpp defaults to using Hunter to download its dependencies off the 'net,
even if they exist locally. This is convenient for development but not
practical for some production environments. It also makes life hard for clients
that want to link to jaeger-cpp since Hunter doesn't install those
dependencies. It's necessary to also use Hunter in apps that use a jaeger-cpp
built this way... and that's not always practical.

Accordingly, add support for finding jaeger-cpp's dependencies via the normal
CMake package discovery mechanism.

A sligtly unsightly hack is required for nlohmann json, because its header
moved from json.hpp to nlohmann/json.hpp in 2.1.0.

This introduces support for building with

   cmake -DHUNTER_ENABLED=0

WIP:

  - nothing packages nlohmann json 2.1.0 yet, 2.0.2 is widespread.
    but jaegertracing's code doesn't appear to support 2.0.2. For now you
    should work around it by installing 2.1.x locally from source.
    (no bug opened yet)

  - will fail to compile tests if the local thrift is 0.9.1 since there are
    committed generated files from 0.9.2 (#45) and there's no mechanism
    to regenerate them.

The nlohmann json issue is beyond my C++-fu, but it's easy enough to install 2.1.x and it's just a header. A pain, but manageable.

The Thrift issue #45 is a bit more of an issue; I'm on Fedora 25 with 0.9.1 right now, but Fedora Extras for RHEL7 also carries 0.9.1. More to the point, it's probably bad to rely on a code generator's output being compatible across multiple versions of its library.


ringerc:
Hrm, would also have to add support for non-hunter Travis builds before mergeable.

CMakeLists.txt
```diff
@@ -59,12 +59,12 @@ list(APPEND package_deps thrift)
 hunter_add_package(opentracing-cpp)
 # Not `${hunter_config}` because OpenTracing provides its own
 # OpenTracingConfig.cmake file
-find_package(OpenTracing CONFIG REQUIRED)
-list(APPEND LIBS OpenTracing::opentracing-static)
+find_package(OpenTracing ${hunter_config} REQUIRED)
```
isaachier: Good idea!

CMakeLists.txt
```diff
@@ -283,7 +287,7 @@ if(BUILD_TESTING)
       GTEST_HAS_TR1_TUPLE=0
       GTEST_USE_OWN_TR1_TUPLE=0)
   target_link_libraries(
-      UnitTest testutils GTest::main jaegertracing-static ${LIBS})
+      UnitTest testutils ${GTEST_MAIN_LIBRARY} jaegertracing-static ${LIBS})
```
isaachier: Not sure if this is compatible with Hunter. May need to do something like this:

```
if(HUNTER_ENABLED)
  set(gtest_lib GTest::main)
else()
  set(gtest_lib ${GTEST_MAIN_LIBRARY})
endif()
```

src/jaegertracing/baggage/RemoteRestrictionJSON.h
```diff
@@ -18,10 +18,17 @@
...
 #include <string>
 #include <vector>
 
+#include "jaegertracing/Constants.h"
+/* nlohmann json changed locations (in 2.1.0?) */
+#ifdef JAEGERTRACING_NLOHMANN_UNQUALIFIED_PATH
+  #include <json.hpp>
+#else
+  #include <nlohmann/json.hpp>
+#endif
```
isaachier: Interesting. Another possibility is including the header somewhere because I think it is a header-only library.

src/jaegertracing/testutils/CMakeLists.txt
```diff
@@ -6,4 +6,9 @@ TUDPTransport.cpp)
 add_library(testutils ${SRC})
 target_include_directories(testutils PUBLIC
     $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src>)
-target_link_libraries(testutils thrift::thrift_static)
+if(HUNTER_ENABLED)
+  target_link_libraries(testutils thrift::thrift_static)
+else()
+  target_link_libraries(testutils ${THRIFT_LIBRARIES})
+endif()
```
isaachier: Nice

ringerc:
No major changes in this push, no need to re-examine.

Will try to address the Thrift stuff but am *rapidly* running out of time for what was going to be a 1-day prototype. I can't even link it to postgres yet :( so may have to abandon.

ringerc:
Also consider that opentracing-cpp does not its self use Hunter, so it's a
requirement introduced by jaeger cpp-client.

ringerc:
Pushed new commits. In particular I've now added the docs on top, since they didn't make much sense without local dependencies support.

I don't think I can get Travis support in too simply because of the quirky dependencies on Thrift and json-cpp. But anyway, it works...

src/jaegertracing/baggage/RemoteRestrictionJSON.h
ringerc: Removed from latest push since cpp-client doesn't build against older headers due to code compat issues anyway (#47)

ringerc:
I landed up mostly rewriting the build system to make this work fairly cleanly. It now follows https://rix0r.nl/blog/2015/08/13/cmake-guide/ practices: modular `CMakeLists.txt`, uses `IMPORT` targets, etc. You'll find it a lot easier to follow and work with, and it doesn't have to special case stuff for Hunter anymore.

ringerc:
I also got rid of the testutils subproject, instead building them as part of the test executables.

I haven't looked closely at how to do a local-dependencies build in Travis yet. The usual technique would be listing package dependencies but the very specific Thrift dependency and very new json dependency make that impractical. 

I think it's ready to merge. I rolled the README into it too btw.

ringerc:
Hrm, the unit tests fail for some reason...

ringerc:
Test failures were due to finding the installed headers. Fixed.

ringerc:
@isaachier OK, think we're good to go :)

CMakeLists.txt
```diff
@@ -33,40 +47,29 @@ else()
...
-list(APPEND package_deps Boost)
 
 hunter_add_package(thrift)
-find_package(thrift ${hunter_config} REQUIRED)
-if(HUNTER_ENABLED)
-  list(APPEND LIBS thrift::thrift_static)
-else()
-  list(APPEND LIBS ${THRIFT_LIBRARIES})
+find_package(thrift 0.9.2 ${hunter_config} REQUIRED)
+if(thrift_VERSION VERSION_GREATER "0.9.3")
```
isaachier: Maybe add `OR VERSION_LESS "0.9.2"`.

CMakeLists.txt
```diff
@@ -14,7 +17,18 @@ HunterGate(
     LOCAL
 )
 
-project(jaegertracing VERSION 0.3.0)
+project(jaegertracing VERSION 0.3.0 LANGUAGES CXX)
+
+include(GNUInstallDirs)
```
isaachier: Is this used?

CMakeLists.txt
```diff
@@ -33,40 +47,29 @@ else()
   set(hunter_config "")
 endif()
 
-set(package_deps)
-
 if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND
     CMAKE_CXX_COMPILER_VERSION VERSION_LESS "4.9")
   set(boost_components regex)
   hunter_add_package(Boost COMPONENTS regex)
-  list(APPEND LIBS Boost::regex)
```
isaachier: Does this break the Hunter build?

CMakeLists.txt
```diff
@@ -329,33 +114,6 @@ configure_package_config_file(
...
-#   * header location after install: <prefix>/include/jaegertracing/Tracer.h
-#   * headers can be included by C++ code `#include <jaegertracing/Tracer.h>`
-install(
-  TARGETS jaegertracing jaegertracing-static
-  EXPORT "${TARGETS_EXPORT_NAME}"
-  LIBRARY DESTINATION "lib"
-  ARCHIVE DESTINATION "lib"
-  RUNTIME DESTINATION "bin"
-  INCLUDES DESTINATION "${include_install_dir}"
-)
```
isaachier: Why remove this?

CMakeLists.txt
```diff
@@ -371,3 +129,12 @@ install(
   NAMESPACE "${namespace}"
   DESTINATION "${config_install_dir}"
 )
+
+option(BUILD_TESTING "Build unit test" ON)
```
isaachier: This option is built-in to CMake. No need to redeclare.

CMakeLists.txt
```diff
@@ -371,3 +129,12 @@ install(
...
+if (BUILD_TESTING)
+  enable_testing()
+  add_custom_target(check COMMAND echo "run make test instead")
```
isaachier: Cool

README.md
```diff
@@ -3,6 +3,228 @@
...
+Java and Go.
+
+## Quick build
+
+To do a test build and install in $HOME/jaeger-cpp-client:
+
+    git clone https://github.com/jaegertracing/cpp-client.git jaeger-cpp-client
+    mkdir jaeger-cpp-build
+    cd jaeger-cpp-build
+    cmake -DCMAKE_INSTALL_PREFIX=$HOME/jaeger-cpp-client ../jaeger-cpp-client
```
isaachier: This command should still work with existing install directives.

README.md
```diff
@@ -3,6 +3,228 @@
...
+            )cfg";
+
+        const auto config = jaegertracing::Config::parse(YAML::Load(kConfigYAML));
+        auto tracer = jaegertracing::Tracer::make("postgresql", config);
+        opentracing::Tracer::InitGlobal(tracer);
+    }
+
+    void teardownTracing()
+    {
+        opentracing::Tracer::Global()->Close
```
isaachier: Does it need to be closed?

README.md
```diff
@@ -3,6 +3,228 @@
...
+The Jaeger tracing instance usually lives as long as the traced process.
+
+The tracer should be closed before program exit so it can flush its buffers,
+so traces aren't lost.
+
+### Usage from C
+
+It's possible to use `jaeger-cpp` from C with appropriate `extern "C"` thunks
+and care around resource lifetime management etc. A real world example is the
+support for opentracing and jaeger in ngnix; see
```
isaachier: Typo: `nginx`

src/jaegertracing/CMakeLists.txt
```diff
@@ -0,0 +1,195 @@
+# Build the jaegertracing library its self
+cmake_minimum_required(VERSION 3.3)
+
+project(libjaegertracing VERSION ${jaegertracing_VERSION} LANGUAGES CXX)
```
isaachier: This seems like a bad idea. It should all be one project. Not sure why splitting at all TBH.

ringerc:
@isaachier It passes Hunter builds in CI and on my local machine. I spent a fair while preserving that and making everything uniform.

CMakeLists.txt
ringerc: That's handled by `find_package`. It doesn't have good support for saying "these two versions ONLY" hence the warning.

CMakeLists.txt
ringerc: No, good catch.

CMakeLists.txt
ringerc: No, if there's a library to link assocated with `Boost::regex` (for old compilers where it's not header-only, apparently) it's specified using `target_link_libraries` for `jaegertracing`.

Though come to think of it, that should be unconditional. On newer g++ the `Boost::regex` target will be an `INTERFACE` library instead.

CMakeLists.txt
ringerc: Move, not remove. Installation of the library target is in the library's own `CMakeLists.txt` in `src/jaegertracing`.

CMakeLists.txt
ringerc: Thanks. My last significant CMake experience was a *long* time ago.

README.md
ringerc: I don't understand

README.md
ringerc: AFAICS if you don't close the tracer you might not flush all spans. But I guess it depends on the implementation details I haven't been able to dig into yet about whether any caching and delays are present for sending span closes.

I'd argue it's good practice; if you don't have to now, you probably will later when it starts buffering spans and sending them in chunks to reduce syscall overheads, etc.

README.md
ringerc: Will fix

src/jaegertracing/CMakeLists.txt
ringerc: Separate project: convenience and seems to be standard practice with cmake, but I don't feel strongly about it.

Splitting at all: readability, maintainability, separation of concerns. The tests don't intermix with the main library and trample on its settings, etc.

See also https://rix0r.nl/blog/2015/08/13/cmake-guide/ and many others.

If I land up adopting this seriously I'll probably send a followup pull to extract the test sources into a separate subtree and ensure there's zero inclusion of test sources by non-test code, really make the boundary clear. I'll also submit another to pull the thrift generated code out of the source tree and build it on demand, something that'll be much easier with the possible fix to #45 I've been looking at.

CMakeLists.txt
ringerc: I'm not 100% sure why the build uses `hunter_add_package(Boost COMPONENTS regex)` for older gcc, and simply `hunter_add_package(Boost)` for newer. That seems ... weird. Wouldn't you just always ask for `Boost::regex` with `COMPONENTS regex` ?

ringerc:
@isaachier Pushed with comments addressed. I left the subproject in place for now; personally I think it's good to be able to refer to parts of the codebase by name etc. But not wedded to that. Removing it just requires changing some paths and vars a bit.

It now unconditionally uses `Boost::regex`. That works fine with/without Hunter, and whether Boost::regex is a header-only lib or requires linkage.

CMakeLists.txt
isaachier: Right. I guess my point is that `find_package` works differently with Hunter vs. non-Hunter.

Here's a brief overview. CMake offers two ways of finding packages (and maybe a third but IDK about it): modules and configs. Modules refer to the `Find<Package>` modules that you added here and that many people use in CMake builds. The problem is inherent redundancy of many projects. Why does everyone need to add these modules? So CMake offered another approach whereby the package declares its location during its CMake build, then other packages can easily locate it. The latter method is called the config approach.

Hunter relies SOLELY on the config approach. That means none of your `Find<Package>` files will be called at all. Relying on any code in there isn't going to work.

README.md
isaachier: It seems you are correct:

```
  // Close is called when a tracer is finished processing spans. It is not
  // required to be called and its effect is unspecified. For example, an
  // implementation might use this function to flush buffered spans to its
  // recording system and failing to call it could result in some spans being
  // dropped.
```
https://github.com/opentracing/opentracing-cpp/blob/master/include/opentracing/tracer.h#L147

src/jaegertracing/CMakeLists.txt
isaachier: I have seen splitting the `CMakeLists.txt` file into multiple subdirectories, but I guess not subprojects. I see it is allowed in the link you pointed out, just hope it doesn't mess with other dependencies, etc.

CMakeLists.txt
ringerc: Yeah, I realise that. That's why `${hunter_config}` substitutes the `CONFIG` option to `find_package`.

Have a look at the CMake package config files CMake generates from the Hunter-built packages. You'll see that they `add_library(blah SHARED IMPORT)` and setting `INTERFACE` properties like `INTERFACE_INCLUDE_DIRECTORIES` and `INTERFACE_INCLUDE_DIRECTORIES` on the resulting target. That's exactly what I'm doing in the `Find` packages too, to be consistent with CMake `CONFIG` mode. Look at the bundled CMake modules in `/usr/share/cmake/Modules` (or wherever yours are installed) e.g. `FindGTest.cmake`. You'll find that they *also* declare import targets and set properties on them for the library name, header paths, etc.

So it's not really to do with Hunter or non-Hunter. There are two things at play here:

* Is it using `CONFIG` mode (as Hunter requests), `MODULE` mode where it only uses `Find` packages, or the default where it tries a module first then falls back to config-mode?

* Does the resulting configuration created by a CONFIG file load or a Find module execution generate import-targets (new-style CMake), or just set variables (old-style CMake)?

What I've done is written the Find modules to produce the same import targets, etc, as what you get from loading the `CONFIG` module. So it all "Just Works". It's done how CMake's own bundled modules do it.

None of this is helped by the fairly erratic CMake documentation, but have a look at:

* https://cmake.org/cmake/help/v3.3/command/add_library.html#imported-libraries
* https://cmake.org/cmake/help/v3.3/prop_tgt/IMPORTED.html#prop_tgt:IMPORTED
* https://cmake.org/cmake/help/v3.3/manual/cmake-packages.7.html
* https://cmake.org/cmake/help/v3.0/prop_tgt/INTERFACE_INCLUDE_DIRECTORIES.html
* https://cmake.org/cmake/help/v3.0/prop_tgt/INTERFACE_LINK_LIBRARIES.html
* https://cmake.org/cmake/help/v3.0/prop_tgt/INTERFACE_COMPILE_DEFINITIONS.html

CMake takes care of the dependency chains, adding link libraries and include paths when you reference an import target, etc. It uses the properties set on the target for this. There's no need for a `LIBS` variable or any of that anymore.

This is also why the build-tree passed to headers in `target_include_directories` are `PRIVATE` and the install-path headers are `INTERFACE`. This tells CMake not to include the build-tree headers in the CMake config file it generates. And to add the install-target headers in the config file, but not use them when compiling the package its self, since it won't be installed yet.

CMake really needs a docs section that properly explains `PRIVATE`, `PUBLIC` and `INTERFACE`, explains `IMPORT` targets and how they're used in `Find` modules, etc. It's there, you just have to find all the scattered pieces from the docs and work out the whole picture from it. Dealing with this has been a massive pain, but hopefully it's nearly done.

src/jaegertracing/CMakeLists.txt
ringerc: OK, I'll remove the subproject

ringerc:
Subproject definitions removed. Also added detection of Thrift 0.11.0 (though not build support for it) per #45

ringerc:
@isaachier Any thoughts?

isaachier:
Sorry for delay here. I'd like to try this along with the changes in #48. Can you copy the modified Travis files there to allow for testing with and without Hunter?

ringerc:
@isaachier I merged the Travis files as requested

CMakeLists.txt
isaachier: So the reason for that is the header-only vs. library dependencies in Boost. Hunter needs to know which libraries it is going to build. I'd rather not build or use the regex library from Boost if standard C++11 regex is available.

CMakeLists.txt
isaachier: OK

CMakeLists.txt
isaachier: No worries :)

README.md
isaachier: I just mean this seems to imply the user must specify a prefix. It is optional, not a must.

src/jaegertracing/CMakeLists.txt
isaachier: Thanks

isaachier:
I owe you a full review soon. Will get around to it ASAP.

ringerc:
On 14 February 2018 at 06:37, isaachier <notifications@github.com> wrote:

> *@isaachier* commented on this pull request.
> ------------------------------
>
> In CMakeLists.txt
> <https://github.com/jaegertracing/cpp-client/pull/46#discussion_r168025189>
> :
>
> >  if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND
>      CMAKE_CXX_COMPILER_VERSION VERSION_LESS "4.9")
>    set(boost_components regex)
>    hunter_add_package(Boost COMPONENTS regex)
> -  list(APPEND LIBS Boost::regex)
>
> So the reason for that is the header-only vs. library dependencies in
> Boost. Hunter needs to know which libraries it is going to build. I'd
> rather not build or use the regex library from Boost if standard C++11
> regex is available.
>

Ah, I misunderstood it then.

I read it as something like "on gcc < 4.9 we need to link to something to
use Boost::regex; on later versions we can use it as a header-only
library". Reading comprehension fail.

Ah, hence regex_namespace in utils/Regex.h .

I'll rewrite that a little if you're OK with it, so that CMake injects a
macro into Constants.h to set JAEGERTRACING_USE_STD_REGEX rather than
making the decision in  utils/Regex.h. That way the intent should be
clearer.

-- 
 Craig Ringer                   http://www.2ndQuadrant.com/
 PostgreSQL Development, 24x7 Support, Training & Services

isaachier:
That makes sense, but the issue is that a team at Uber uses [Bazel](https://bazel.build) for their build, not CMake, and the fix is primarily for them. Not sure if they can use a CMake based macro. Then again, I think they hack a lot of it at the moment to get it working so not a huge deal.

ringerc:
On 14 February 2018 at 10:15, isaachier <notifications@github.com> wrote:

> That makes sense, but the issue is that a team at Uber uses Bazel
> <https://bazel.build> for their build, not CMake, and the fix is
> primarily for them. Not sure if they can use a CMake based macro. Then
> again, I think they hack a lot of it at the moment to get it working so not
> a huge deal.
>
They wouldn't need to, it'd be injected into Config.h . Unless you mean
that they build jaegertracing its self with Bazel, in which case they're
already duplicating the whole build system.

PostgreSQL uses autotools (sigh), so I'm not exactly keen to add hard CMake
dependencies myself.

-- 
 Craig Ringer                   http://www.2ndQuadrant.com/
 PostgreSQL Development, 24x7 Support, Training & Services

isaachier:
Right lol I get that. Ironically, Hunter supports autotools and maintains a PostgreSQL build, so you could probably leverage that somehow in a worst case scenario. Either way, my support for GCC 4.8 and below is declining. nlohmann specifically won't work on 4.8. There are some patches the team I work with used but I'd recommend not using them.

ringerc:
On 14 February 2018 at 11:56, isaachier <notifications@github.com> wrote:

> Right lol I get that. Ironically, Hunter supports autotools and maintains
> a PostgreSQL build, so you could probably leverage that somehow in a worst
> case scenario. Either way, my support for GCC 4.8 and below is declining.
> nlohmann specifically won't work on 4.8. There are some patches the team I
> work with used but I'd recommend not using them.
>
Yeah. 4.8 isn't so old, though; notably RHEL 7 has 4.8 and RHEL6 has 4.4.

That means I won't be able to deploy tracing on pretty much any customer
system.

I think I have to drop this. Between the Thrift issues,  nlohmann version,
etc, and the implied gcc dependency, I can't actually *use it* anywhere.

-- 
 Craig Ringer                   http://www.2ndQuadrant.com/
 PostgreSQL Development, 24x7 Support, Training & Services

isaachier:
Ya I'm sorry again and not entirely surprised. This library was built for a very modern C++ application. That's why I've been working on a C version - - to reduce dependencies and increase portability.

ringerc:
Thanks, and sorry I'm not actively contributing to the C one yet. I've
burned all my time and then some trying to fix the C++ one up for my needs
/ for general portability.

I can hack around with a specific Thrift version + installing a newer gcc
or icc or clang, albeit painfully. Most likely I'd build packages and ship
them with a new libstdc++. In the mean time it remains useful for
development & debugging. I'll try to get work to dedicate some
time/resources to the C backend but unsure how I'll go.

isaachier:
@ringerc no worries. Eventually, I hope to remove much of the C++ internals from the C++ client, wrapping the C core instead. In that case, your code would work fine, even if not using C directly. When it gets to that level of maturity, I'll be sure to let you know.

ringerc:
Good to know.

I'll try to pick up the C client and do some testing in a bit. I'd rather
adopt it than wrap the C client in a C++ API then wrap the C++ API in a C
one for the app.

PR #50, opened 2018-01-31
Document that the Thrift generated files are patched

ringerc:
For #45, document that the code-gen output from the IDL is currently patched from the original generated code, and why.


yurishkuro:
@ringerc could you please amend the commit to make it signed? `git commit --amend -s`

ringerc:
@yurishkuro Done

isaachier:
Awesome work! Thank you for the docs.

Issue #53, opened 2018-02-05
Expose interface to flush buffered spans


ringerc:
The tracer buffers closed span, only writing them out intermittently.

Sometimes the client will know it has something to send and want to flush it promptly.

I suggest extending the OpenTracing interface to expose an explicit Flush on jaegertracing::Tracer.

Happy to send a patch if you think it's reasonable.

ringerc:
Also requires promoting `RemoteReporter::flush()` to be a virtual method of `Reporter`.

tedsuo:
Hi @ringerc, changes to the OpenTracing interface should be made against https://github.com/opentracing/opentracing-cpp instead of jaeger.

But it sounds like the issue is that the jaeger client does not flush spans on close (which it should), not that the flush method isn't directly exposed to 3rd party instrumentation.

ringerc:
That's ​correct. However, in the course of working on that I noticed that
there were places in my app where I'd really like to flush trace
information eagerly. For example, at the completion of some top-level
request handler where I *know* there's work to send, and won't be work
afterwards.

Ideally I'd not have timers on every backend ticking away anyway; that's
going to get expensive in a multi-process server like postgresql. Being
able to explicitly flush so I can do it from postgres's own event handling
would be a big help. Hence this open issue.

rnburn:
@ringerc -- that strategy sounds similar to how envoy does tracing
https://github.com/envoyproxy/envoy/blob/master/source/common/tracing/lightstep_tracer_impl.cc#L106
I made an interface for lightstep's tracer that allows it you used in that manner (no threading, or sockets managed by the tracer); but didn't pursue standardizing the interface as I think it would be harder to make a vendor-neutral API from it.

Would be interested in getting some performance data to see what the cost of tracer's doing their own threading and timers is.

isaachier:
Interesting idea @ringerc, maybe I will reopen #54 after all...

ringerc:
The reason I care somewhat about the threading and timers is that I'm
working on a multi-processing shared-nothing-by-default system where
everything is fork()ed.

So there are lots of procs in a service, each with their own tracer
instance.

Those timers can add up. As can the threads.

It's not a big concern at this point, and if it becomes one I'll submit a
separate patch that lets Jaeger cpp-client entirely disable automatic
flushing and use only the main-thread. Shouldn't be too hard (famous last
words).

The reason I want forced flushing now isn't so much overhead management as
[...]

PR #54, opened 2018-02-05
Flush any pending buffers on close()

### LINKED ISSUE #52: cpp-client loses spans on short-lived processes; Close() not flushing buffers?
If a process:

* Creates a tracer with a typical configuration (like the following yml)
* Creates a span
* Finishes the span
* Closes the tracer
* Exits

the span should be reported to the collector. But it isn't unless a short sleep is inserted *before* the `tracer->Close()` call.

If the `sleep()` is added *after*  the tracer close, the spans are not sent. It looks like Close() must be failing to flush buffers, instead dropping any pending buffers on the floor.

Configuration, loaded with the YAML config support:

```
disabled: false
sampler:
    type: const
    param: 1
reporter:
    queueSize: 100
    bufferFlushInterval: 10
    logSpans: false
    localAgentHostPort: 127.0.0.1:6831
headers:
    jaegerDebugHeader: debug-id
    jaegerBaggageHeader: baggage
    TraceContextHeaderName: trace-id
    traceBaggageHeaderPrefix: "testctx-"
baggage_restrictions:
    denyBaggageOnInitializationFailure: false
    hostPort: 127.0.0.1:5778
    refreshInterval: 60
```

ringerc:
The RemoteReporter buffers spans, but didn't flush them on close.
So any spans Finish()ed between the last flush interval and the
Close() of the Tracer would be lost.

Fixes #52

Signed-off-by: Craig Ringer <craig@2ndquadrant.com>

### DIFF
--- a/src/jaegertracing/Tracer.h
+++ b/src/jaegertracing/Tracer.h
@@ -102,7 +102,7 @@ class Tracer : public opentracing::Tracer,
             serviceName, sampler, reporter, logger, metrics, options));
     }
 
-    ~Tracer() { Close(); }
+    ~Tracer() { close(); }
 
     std::unique_ptr<opentracing::Span>
     StartSpanWithOptions(string_view operationName,
@@ -182,6 +182,12 @@ class Tracer : public opentracing::Tracer,
 
     void Close() noexcept override
     {
+        flush();
+        close();
+    }
+
+    void close() noexcept
+    {
         try {
             _reporter->close();
             _sampler->close();
@@ -192,7 +198,10 @@ class Tracer : public opentracing::Tracer,
         }
     }
 
-    void close() noexcept { Close(); }
+    void flush()
+    {
+        _reporter->flush();
+    }
 
     const std::string& serviceName() const { return _serviceName; }
 
--- a/src/jaegertracing/reporters/RemoteReporter.cpp
+++ b/src/jaegertracing/reporters/RemoteReporter.cpp
@@ -39,6 +39,7 @@ RemoteReporter::RemoteReporter(const Clock::duration& bufferFlushInterval,
     , _running(true)
     , _lastFlush(Clock::now())
     , _cv()
+    , _cv_flush()
     , _mutex()
     , _thread()
 {
@@ -94,8 +95,10 @@ void RemoteReporter::sweepQueue()
                 sendSpan(span);
             }
             else if (bufferFlushIntervalExpired()) {
-                flush();
+                async_flush();
             }
+			/* If anyone's waiting on flush completion, notify them */
+            _cv_flush.notify_one();
         } catch (...) {
             auto logger = logging::consoleLogger();
             assert(logger);
@@ -122,12 +125,13 @@ void RemoteReporter::sendSpan(const Span& span)
     }
 }
 
-void RemoteReporter::flush()
+void RemoteReporter::async_flush()
 {
     try {
         const auto flushed = _sender->flush();
         if (flushed > 0) {
             _metrics.reporterSuccess().inc(flushed);
+            _cv.notify_one();
         }
     } catch (const Transport::Exception& ex) {
         _metrics.reporterFailure().inc(ex.numFailed());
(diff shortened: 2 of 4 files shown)


rnburn:
btw - is calling close automatically like this in the reporter's destructor really a good idea?
https://github.com/jaegertracing/cpp-client/blob/master/src/jaegertracing/reporters/RemoteReporter.h#L45
it can cause the process to take much longer to exit than it otherwise would.

ringerc:
On 5 February 2018 at 22:40, Ryan <notifications@github.com> wrote:

> btw - is calling close automatically like this in the reporter's
> destructor really a good idea?
> https://github.com/jaegertracing/cpp-client/blob/master/src/jaegertracing/
> reporters/RemoteReporter.h#L45
> it can cause the process to take much longer to exit than it otherwise
> would.
>

Yeah. Possibly not in the dtor; maybe it's best done only via an explicit
Close() from the Tracer. Need to expose internal API for that.

ringerc:
OK, so addressing #53 instead with an extension `Tracer.Flush()` method to allow explicit flushes.

ringerc:
Hrm, doesn't seem to do the job. I'm still losing spans. General approach seems sensible, forcing a blocking flush via separate API request.  I just haven't figured out how to do it yet through all the layers involved.

ringerc:
Maybe something like

```
 void RemoteReporter::flush()
{
    async_flush();
    std::unique_lock<std::mutex> lock(_mutex);
    _cv.wait(lock, [this]() {
        return !_running || _queue.empty();
    });
}
```

?

isaachier:
Actually, I see the issue is that I didn't add a "poison pill" to the end of the queue, instead used a boolean flag `running`. That means the shutdown is immediate. Will patch and submit new PR. Thanks for the help here, but will be taking a different approach, the one used in the [Go client](https://github.com/jaegertracing/jaeger-client-go/blob/master/reporter.go).

isaachier:
Actually, I see we do check for an empty queue in the thread's main loop, so I'm not sure why this isn't effectively the same thing. Also, now that I've read #53, I see this issue has further reaching goals than just fixing this local problem. Might be worth adding an eager flush mechanism as you suggested.

src/jaegertracing/Tracer.h
```diff
@@ -194,6 +194,11 @@ class Tracer : public opentracing::Tracer,
 
     void close() noexcept { Close(); }
 
+    void Flush()
```
isaachier: I only use the title case functions for opentracing-cpp methods. I'd prefer `flush` here unless it is added to the OT implementation and this is an override.

ringerc: Fixed in latest

ringerc:
I'm still seeing intermittent lost spans with this patch, though less frequently.

I just did some test runs with Wireshark observing the udp communications. The lost spans definitely never seem to get sent on the wire, they're not misplaced in the collector. The span name / operation name never appears on the wire at all.

Cause is uncertain as yet. Having trouble debugging effectively. May have to write small test program to see if I can repro.

ringerc:
OK, modified as discussed. It now flushes buffers on explicit `Close()` only.

My implementation of `RemoteReporter::flush()` was pretty much completely wrong. Fixed now, so it waits on its own condition variable and the reporter thread sets that condition variable after a wakeup (whether or not a span was flushed).

ringerc:
@isaachier Should be ready for review/merge

src/jaegertracing/reporters/RemoteReporter.cpp
```diff
@@ -137,5 +141,14 @@ void RemoteReporter::flush()
     _lastFlush = Clock::now();
 }
 
+void RemoteReporter::flush()
+{
+    std::unique_lock<std::mutex> lock(_mutex);
+    _cv_flush.wait(lock, [this]() {
+        async_flush();
```
isaachier: This definitely looks wrong. The lambda here is just meant to express a boolean condition to avoid spurious wakeups. For example, instead of `while (running) { cv.wait(); }` you can do `cv.wait([&running] { return running; }`. This will potentially lead to many bugs in the program.

isaachier:
Fixed in #59.

ringerc:
Thanks!​

PR #58, opened 2018-02-08
Record logs passed with Span::FinishWithOptions

### LINKED ISSUE #57: Logs supplied in Span::FinishWithOptions are silently discarded
`jaegertracing::Span::FinishWithOptions` entirely ignores the `options.log_records` field of `opentracing::FinishSpanOptions`.

Any logs so passed are discarded.

Not ideal, because the `Span::Log` method only accepts a `std::initializer_list` so it's clumsy when you have a dynamic set of keys in your logs.

Will send a pull, this just documents the issue first.

ringerc:
Jaeger cpp-client was dropping logs included in a FinishSpanOptions
silently on the floor. Append them to the log buffer during FinishWithOptions instead.

Fixes #57

Signed-off-by: Craig Ringer <craig@2ndquadrant.com>

### DIFF
--- a/src/jaegertracing/LogRecord.h
+++ b/src/jaegertracing/LogRecord.h
@@ -25,6 +25,8 @@
 #include <type_traits>
 #include <vector>
 
+#include <opentracing/span.h>
+
 namespace jaegertracing {
 
 class LogRecord {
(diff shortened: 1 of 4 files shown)


ringerc:
This requires OpenTracing master by the looks

rnburn:
This should be ready to merge now.

isaachier:
Rebuilding now. Sorry for delay.

PR #59, opened 2018-02-09
Remote reporter flush fix

isaachier:
Guarantee a flush occurs on destruction of `RemoteReporter`. Add more `noexcept` specifiers for reporter methods.


src/jaegertracing/reporters/RemoteReporter.cpp
```diff
@@ -60,21 +60,25 @@ void RemoteReporter::report(const Span& span)
...
-        if (!_running) {
-            return;
+    try {
+        {
+            std::lock_guard<std::mutex> lock(_mutex);
+            if (!_running) {
+                return;
+            }
+            _running = false;
+            flush();
```
isaachier: Most important line in this review.

ringerc:
Tested and works for me here, passes the shortlived test proc.

Thanks very much.

PR #61, opened 2018-02-16
Support Thrift 0.11.0 and OpenSSL 1.1

ringerc:
Support Thrift 0.11.0 and OpenSSL 1.1

This isn't a pull request for serious consideration. I'm creating it so that other people can find it. This is my integration branch that adds support for Thrift 0.11.0, OpenSSL 1.1, building without Hunter, and logs appended using the Span::FinishWithOptions. It also adds a README.

This is the jaeger `cpp-client` I use and deploy.


ringerc:
Closing; this isn't for serious consideration, only so people can discover it.

PR #64, opened 2018-03-16
Add support for dynamic loading.

### LINKED ISSUE #3: Support dynamic loading
https://github.com/opentracing/opentracing-cpp/issues/28

rnburn:
Implements the dynamic loading support available in opentracing-cpp-1.3.

Fixes #3

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -287,7 +287,7 @@ if(BUILD_TESTING)
       GTEST_HAS_TR1_TUPLE=0
       GTEST_USE_OWN_TR1_TUPLE=0)
   target_link_libraries(
-      UnitTest testutils GTest::main jaegertracing-static ${LIBS})
+      UnitTest testutils GTest::main jaegertracing ${LIBS})
   add_test(NAME UnitTest COMMAND UnitTest)
 
   add_executable(DynamicallyLoadTracerTest src/jaegertracing/DynamicallyLoadTracerTest.cpp)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -262,6 +262,7 @@ if(BUILD_TESTING)
       src/jaegertracing/TraceIDTest.cpp
       src/jaegertracing/TracerTest.cpp
       src/jaegertracing/TracerFactoryTest.cpp
+      src/jaegertracing/DynamicLoadTest.cpp
       src/jaegertracing/UDPTransportTest.cpp
       src/jaegertracing/baggage/BaggageTest.cpp
       src/jaegertracing/metrics/MetricsTest.cpp
@@ -289,10 +290,10 @@ if(BUILD_TESTING)
       UnitTest testutils GTest::main jaegertracing-static ${LIBS})
   add_test(NAME UnitTest COMMAND UnitTest)
 
-  add_executable(DynamicLoadTest src/jaegertracing/DynamicLoadTest.cpp)
+  add_executable(DynamicallyLoadTracerTest src/jaegertracing/DynamicallyLoadTracerTest.cpp)
   target_link_libraries(
-    DynamicLoadTest OpenTracing::opentracing-static)
-  add_test(NAME DynamicLoadTest COMMAND DynamicLoadTest 
+    DynamicallyLoadTracerTest OpenTracing::opentracing-static)
+  add_test(NAME DynamicallyLoadTracerTest COMMAND DynamicallyLoadTracerTest 
            ${CMAKE_BINARY_DIR}/${CMAKE_SHARED_LIBRARY_PREFIX}jaegertracing${CMAKE_SHARED_LIBRARY_SUFFIX})
   if(JAEGERTRACING_COVERAGE)
       setup_target_for_coverage(NAME UnitTestCoverage
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -294,7 +294,7 @@ if(BUILD_TESTING)
   target_link_libraries(
     DynamicallyLoadTracerTest OpenTracing::opentracing-static)
   add_test(NAME DynamicallyLoadTracerTest COMMAND DynamicallyLoadTracerTest 
-           ${CMAKE_BINARY_DIR}/${CMAKE_SHARED_LIBRARY_PREFIX}jaegertracing${CMAKE_SHARED_LIBRARY_SUFFIX})
+    ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_SHARED_LIBRARY_PREFIX}jaegertracing${CMAKE_SHARED_LIBRARY_SUFFIX})
   if(JAEGERTRACING_COVERAGE)
       setup_target_for_coverage(NAME UnitTestCoverage
                                 EXECUTABLE UnitTest
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -142,6 +142,8 @@ set(SRC
     src/jaegertracing/Tag.cpp
     src/jaegertracing/TraceID.cpp
     src/jaegertracing/Tracer.cpp
+    src/jaegertracing/TracerFactory.cpp
+    src/jaegertracing/DynamicLoad.cpp
     src/jaegertracing/Transport.cpp
     src/jaegertracing/UDPTransport.cpp
     src/jaegertracing/baggage/BaggageSetter.cpp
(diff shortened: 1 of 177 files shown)


CMakeLists.txt
```diff
@@ -283,8 +287,14 @@ if(BUILD_TESTING)
...
   target_link_libraries(
-      UnitTest testutils GTest::main jaegertracing-static ${LIBS})
+      UnitTest testutils GTest::main jaegertracing ${LIBS})
   add_test(NAME UnitTest COMMAND UnitTest)
+
+  add_executable(DynamicallyLoadTracerTest src/jaegertracing/DynamicallyLoadTracerTest.cpp)
+  target_link_libraries(
+    DynamicallyLoadTracerTest OpenTracing::opentracing-static)
+  add_test(NAME DynamicallyLoadTracerTest COMMAND DynamicallyLoadTracerTest 
+           ${CMAKE_BINARY_DIR}/${CMAKE_SHARED_LIBRARY_PREFIX}jaegertracing${CMAKE_SHARED_LIBRARY_SUFFIX})
```
isaachier: Can you change `${CMAKE_BINARY_DIR}` with `${CMAKE_CURRENT_BINARY_DIR}`. I think it is better for recursive builds.

src/jaegertracing/DynamicLoad.cpp
```diff
@@ -0,0 +1,26 @@
...
+#include <system_error>
+
+#include <opentracing/dynamic_load.h>
+
+#include "TracerFactory.h"
+#include "jaegertracing/Tracer.h"
+
+int OpenTracingMakeTracerFactory(const char* opentracingVersion,
+                                 const void** errorCategory,
+                                 void** tracerFactory)
```
isaachier: Can you add `assert(tracerFactory != nullptr)`?

src/jaegertracing/DynamicLoad.cpp
```diff
@@ -0,0 +1,26 @@
...
+{
+    if (std::strcmp(opentracingVersion, OPENTRACING_VERSION) != 0) {
```
isaachier: I'm going to assume this means the OpenTracing C++ API version. I see you use the version string to generate it. Are you sure we want to use minor and patch versions, or just major version?

src/jaegertracing/DynamicallyLoadTracerTest.cpp
```diff
@@ -0,0 +1,43 @@
+/*
+ * Copyright (c) 2017 Uber Technologies, Inc.
```
isaachier: 2018

src/jaegertracing/DynamicLoadTest.cpp
```diff
@@ -0,0 +1,47 @@
...
+TEST(DynamicLoad, validVersion)
+{
+    const void* errorCategory = nullptr;
+    void* tracerFactory = nullptr;
+    const auto rcode = OpenTracingMakeTracerFactory(
+        OPENTRACING_VERSION, &errorCategory, &tracerFactory);
+    ASSERT_EQ(rcode, 0);
+    ASSERT_EQ(errorCategory, nullptr);
+    ASSERT_NE(tracerFactory, nullptr);
+    delete static_cast<opentracing::TracerFactory*>(tracerFactory);
```
isaachier: Interesting use of `static_cast`. StackOverflow agrees with you here: https://stackoverflow.com/a/941959/1930331.

src/jaegertracing/TracerFactory.cpp
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright (c) 2017 Uber Technologies, Inc.
```
isaachier: 2018

src/jaegertracing/TracerFactory.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include "TracerFactory.h"
+
+#include "jaegertracing/Tracer.h"
+
+namespace jaegertracing {
+
+opentracing::expected<std::shared_ptr<opentracing::Tracer>>
+TracerFactory::MakeTracer(const char* configuration,
+                          std::string& errorMessage) const noexcept try {
+    YAML::Node yaml;
```
isaachier: I think I have the whole YAML dependency wrapped in a macro at the moment. If so, can you add this to wrap the existing code.

```
#ifdef JAEGERTRACING_USE_YAML_CPP
// existing code
#else
// return dummy value or construct config by hand
#endif  // JAEGERTRACING_USE_YAML_CPP
```

src/jaegertracing/TracerFactory.h
```diff
@@ -0,0 +1,32 @@
+/*
+ * Copyright (c) 2017 Uber Technologies, Inc.
```
isaachier: 2018

src/jaegertracing/TracerTest.cpp
```diff
@@ -122,16 +122,15 @@ TEST(Tracer, testTracer)
...
-    auto tagItr =
-        std::find_if(std::begin(tracer->tags()),
-                     std::end(tracer->tags()),
-                     [](const Tag& tag) {
-                         return tag.key() == kJaegerClientVersionTagKey;
-                     });
+    auto tagItr = std::find_if(
+        std::begin(tracer->tags()),
+        std::end(tracer->tags()),
+        [](const Tag& tag) { return tag.key() == kJaegerClientVersionTagKey; });
```
isaachier: Looks like you ran `clang-format` here. Please run `scripts/clang-format.sh` before pushing next commit to make sure all files are formatted correctly.

CMakeLists.txt
rnburn: Done.

src/jaegertracing/DynamicLoad.cpp
rnburn: Added.

src/jaegertracing/DynamicLoad.cpp
rnburn: That might make sense.... though it's pretty easy to break the ABI (for example, if you were to change the order of function in the source your vtable entrys could change). It's probably safest (and also easy) to start by requiring a full match.

src/jaegertracing/DynamicallyLoadTracerTest.cpp
rnburn: Replaced everywhere.

src/jaegertracing/TracerTest.cpp
rnburn: Done.

src/jaegertracing/TracerFactory.cpp
rnburn: I set up dynamic loading to fail if YAML support isn't compiled in and added some test coverage for it as well.

rnburn:
@isaachier - This should be ready for another round of reviewing.

isaachier:
Sorry totally lost track of this. Do you mind changing the copyright from `2018` to `2017-2018` in the files you touched? Otherwise, It looks pretty good to me.

isaachier:
Also, if it isn't too difficult, can you improve coverage of `TracerFactory` by adding an example of a bad config.

rnburn:
@isaachier - There's quite a few tests for invalid configs: https://github.com/rnburn/cpp-client/blob/79b7f6f73091ee8d297b1ae446707ff74e449fcd/src/jaegertracing/TracerFactoryTest.cpp#L22 ... I don't know why they don't seem to be showing up in the coverage.

isaachier:
Hm I'll take a look at build and see whats up

rnburn:
@isaachier replaced 2018 with 2017-2018.

src/jaegertracing/DynamicallyLoadTracerTest.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include <string>
+
+#include <opentracing/dynamic_load.h>
+
+// Verify that Jaeger's shared library can be dynamically loaded and used
+// as a plugin.
+int main(int argc, char* argv[])
+{
+    if (argc != 2) {
+        std::cerr << "DynamicLoadTest <tracer-librar>\n";
```
isaachier: Should be `<tracer-library>`

rnburn: Fixed.

src/jaegertracing/LogRecord.cpp
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017 Uber Technologies, Inc.
+ * Copyright (c) 2017-2018 Uber Technologies, Inc.
```
isaachier: I'm sorry to be annoying about this. The idea was to only change the modified files. That way we only update the changed files. See this commit for a related change in the main Jaeger repo: https://github.com/jaegertracing/jaeger/commit/d7d1c74fa74f77d2c7faa50a0c568f979d9a30fd.

rnburn: Is this what you're looking for? The files I added have 2018 and the file that got modified has 2017-2018.

PR #83, opened 2018-04-15
Add lock to SpanContext to protect baggage

isaachier:
Signed-off-by: Isaac Hier <isaachier@gmail.com>

### DIFF
--- a/src/jaegertracing/SpanContext.h
+++ b/src/jaegertracing/SpanContext.h
@@ -19,6 +19,7 @@
 
 #include <iomanip>
 #include <iostream>
+#include <mutex>
 #include <string>
 #include <unordered_map>
 
@@ -41,6 +42,7 @@ class SpanContext : public opentracing::SpanContext {
         , _spanID(0)
         , _parentID(0)
         , _flags(0)
+        , _mutex()
     {
     }
 
@@ -56,6 +58,7 @@ class SpanContext : public opentracing::SpanContext {
         , _flags(flags)
         , _baggage(baggage)
         , _debugID(debugID)
+        , _mutex()
     {
     }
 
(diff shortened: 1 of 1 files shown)


isaachier:
@rnburn could you please take a look at this and see if it looks OK? @yurishkuro this is the change for C++ client to lock baggage in span context to avoid thread-unsafe return-by-reference semantics in OpenTracing C++ (https://github.com/opentracing/opentracing-cpp/issues/74).

isaachier:
@black-adder or @vprithvi do you mind glancing over this? The background isn't so important just I had to make SpanContext protect baggage with a mutex.

src/jaegertracing/SpanContext.h
```diff
@@ -160,9 +166,17 @@ class SpanContext : public opentracing::SpanContext {
 
     bool operator==(const SpanContext& rhs) const
     {
+        {
+            std::lock(_mutex, rhs._mutex);
+            std::lock_guard<std::mutex> lock(_mutex, std::adopt_lock);
```
black-adder: can this deadlock? this grabs its own lock and rhs grabs its own lock and then we're stuck?

isaachier: I was hoping you wouldn't pay too much attention to this. `std::lock` is a cool trick in C++11 that locks two mutexes without causing deadlock (essentially a loop of `try_lock` under the hood):

> Locks the given Lockable objects lock1, lock2, ..., lockn using a deadlock avoidance algorithm to avoid deadlock.
http://en.cppreference.com/w/cpp/thread/lock

 The lock guard stuff is just to make sure we unlock, doesn't acquire the lock again. Serves same purpose as `defer mutex.Unlock()` would in Go.

isaachier:
Thanks @black-adder

PR #101, opened 2018-05-09
Example app

isaachier:


examples/App.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+}  // anonymous namespace
+
+int main(int argc, char* argv[])
+{
+    if (argc < 2) {
+        std::cerr << "usage: " << argv[0] << " <config-yaml-path>\n";
+        return 1;
+    }
+    setUpTracer(argv[1]);
```
yurishkuro: don't you need to close/flush the tracer before exit?

isaachier: Luckily, no. At least not for the Jaeger client. My favorite thing about C++ is that most things are auto-closing (formally called RAII). Spans must be auto-closing according to OT C++ spec, but the `Tracer::Close()` definition is a little more vague (see https://github.com/opentracing/opentracing-cpp/blob/master/include/opentracing/tracer.h#L149-L153). As of now, we aren't using close any differently here than for spans. Might add it just in case I change my mind.

isaachier: Just to clarify, when the main function exits, the global tracer shared pointer is destroyed, invoking the Tracer destructor, which invokes `Tracer::close` on its own. We could manually invoke any of these methods, but there is no need to.

yurishkuro: Not saying you are wrong, but I'd expect that behavior if you had a locally scoped variable that would get destroyed upon exiting the main scope. That global variables behave the same way is a surprise to me.

isaachier: Ya it's a bit confusing. Here's a source quoting the standard about it: https://stackoverflow.com/a/2204628/1930331.

isaachier:
This works now so landing it.

PR #103, opened 2018-05-10
Fix name of OpenTracing package in CMake config

isaachier:
Signed-off-by: Isaac Hier <isaachier@gmail.com>


isaachier:
Minor fix, just accepting myself.

Issue #107, opened 2018-05-15
tracer.Inject(SpanContext&, HTTPHeadersWriter&) does not use HTTPHeaderPropagator


HugoMFernandes:
`tracer.Inject(SpanContext&, HTTPHeadersWriter&)` is currently using `TextMapPropagator` instead of `HTTPHeaderPropagator`.

Here's the bug: https://github.com/jaegertracing/jaeger-client-cpp/blob/75191fc82acd616ed3c5ed525afefcbf8eed45a1/src/jaegertracing/Tracer.h#L151

I'm guessing this was a copy/paste problem (`Tracer::Extract` is fine). I'll create a PR with the fix in a second.

isaachier:
Fixed in #108.

PR #108, opened 2018-05-15
Fix tracer.Inject(..., HTTPHeadersWriter&) (#107)

HugoMFernandes:
tracer.Inject(SpanContext&, HTTPHeadersWriter&) was using the underlying
TextMapPropagator (instead of HTTPHeaderPropagator). This caused
injected baggage values not to be encoded.

Signed-off-by: Hugo Fernandes <hugof225@gmail.com>


isaachier:
Good catch!

PR #115, opened 2018-06-30
Resolve #112 Build Under Windows

### LINKED ISSUE #112: Make Jaeger Client Cpp build and run on Windows
Make the project available on windows. 

Currently, some header files, some cmake commands and a small part of the networking code make the build fail on Windows and MSVC 2017

mdouaihy:
- Resolves  #112 

* Move to OpenTracing 1.5.0 that builds under Windows
* Change link of OpenTracing to dynamic. With Static linking of OpenTracing, we end up with multi definition of OT Symbols. This is not acceptable especially for the OT Global Tracer
* Fix OS specificities
** Networking API
** Windows headers
* Fix some tests with random behavior that failed on Windows
* Fix some MSVC specific compilation
* Add appveyor build file.

Signed-off-by: Mehrez Douaihy <mehrez.douaihy@gmail.com>


CMakeLists.txt
```diff
@@ -51,11 +60,12 @@ else()
 endif()
 list(APPEND package_deps thrift)
 
+
 hunter_add_package(opentracing-cpp)
 # Not `${hunter_config}` because OpenTracing provides its own
 # OpenTracingConfig.cmake file
 find_package(OpenTracing CONFIG REQUIRED)
-list(APPEND LIBS OpenTracing::opentracing-static)
+list(APPEND LIBS OpenTracing::opentracing)
```
rnburn: Could OpenTracing static linking be provided as an option? It's necessary to make the plugin work with projects that are themselves statically linked or in environments without the c++ standard library installed.

See https://stackoverflow.com/q/47841812/4447365 and https://github.com/envoyproxy/envoy/pull/2252

mdouaihy: The problem I faced with the static link to Open Tracing was that the OT Global Tracer was defined many times. For example, it was defined in the unit test executable and in the Jaeger library.

I can change the code to keep the link as before on Linux and the dynamic link on Windows.

What do you think?

rnburn: I'd be fine with that.

I think there are scenarios where you'd want to link dynamically on linux also, but up to @isaachier how he'd like to support them.

isaachier: Hey I must have totally missed this PR. Will look ASAP.

isaachier: Ya ideally use `MSVC` or the like in `CMakeLists.txt` to make this decision.

CMakeLists.txt
```diff
@@ -208,13 +219,35 @@ cmake_dependent_option(BUILD_SHARED_LIBS "Build shared libraries" ON
...
+endif()
   add_lib_deps(jaegertracing)
   set_target_properties(jaegertracing PROPERTIES
     VERSION ${PROJECT_VERSION}
     SOVERSION ${PROJECT_VERSION_MAJOR})
   set(JAEGERTRACING_LIB jaegertracing)
   list(APPEND JAEGERTRACING_LIBS jaegertracing)
+
+  if (MSVC)
+    # Copy the shared opentracing shared library next to jaeger tracing shared library
```
isaachier: I understand the reason for this, but IDK if this is the ideal solution. Please ask @ruslo in the Hunter gitter chat.

CMakeLists.txt
```diff
@@ -244,17 +284,21 @@ configure_file(
 
 if(JAEGERTRACING_BUILD_EXAMPLES)
   add_executable(app examples/App.cpp)
-  target_link_libraries(app PUBLIC ${JAEGERTRACING_LIB})
+  target_link_libraries(app PUBLIC jaegertracing)
```
isaachier: Please undo this change.

CMakeLists.txt
```diff
@@ -244,17 +284,21 @@ configure_file(
...
 endif()
 
 if(BUILD_TESTING)
+
+  set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
```
isaachier: Again, unsure if this is ideal. @ruslo should have some ideas here too.

CMakeLists.txt
```diff
@@ -289,18 +333,20 @@ if(BUILD_TESTING)
...
   if(TARGET jaegertracing)
     add_executable(DynamicallyLoadTracerTest
       src/jaegertracing/DynamicallyLoadTracerTest.cpp)
+
     target_include_directories(DynamicallyLoadTracerTest PUBLIC
       $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src>
       $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/src>)
     target_link_libraries(
-      DynamicallyLoadTracerTest OpenTracing::opentracing-static GTest::main)
+      DynamicallyLoadTracerTest OpenTracing::opentracing GTest::main)
```
isaachier: Please revert for non-Windows. I remember specifically having issues with this on Mac and/or Linux.

cmake/Hunter/config.cmake
```diff
@@ -0,0 +1,2 @@
+
+hunter_config(opentracing-cpp VERSION ${HUNTER_opentracing-cpp_VERSION} CMAKE_ARGS BUILD_SHARED_LIBS=ON)
```
isaachier: Again, don't believe this is ideal. Hunter already has `HUNTER_BUILD_SHARED_LIBS` to control this. Please remove.

src/jaegertracing/Compilers.h
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright (c) 2017 Uber Technologies, Inc.
```
isaachier: Wrong year.

src/jaegertracing/Compilers.h
```diff
@@ -0,0 +1,31 @@
...
+
+#ifndef JAEGERTRACING_COMPILERS_H
+#define JAEGERTRACING_COMPILERS_H
+
+#ifdef _MSC_VER
+
+#pragma warning(push)
+#pragma warning(disable : 4251)
+
+// Define NOMINMAX to inhibit definition of Macros min(a,b) and max(a,b) in
```
isaachier: Interesting fix.

src/jaegertracing/DynamicLoad.cpp
```diff
@@ -19,31 +19,36 @@
...
 #include <opentracing/dynamic_load.h>
 
+#include "DynamicLoad.h"
 #include "TracerFactory.h"
 #include "jaegertracing/Tracer.h"
 
-int OpenTracingMakeTracerFactory(const char* opentracingVersion,
-                                 const void** errorCategory,
-                                 void** tracerFactory)
+int JaegerTracingMakeTracerFactoryFct(const char* opentracing_version,
```
isaachier: Why?

src/jaegertracing/DynamicLoad.h
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright (c) 2017 Uber Technologies, Inc.
```
isaachier: Year issues again.

src/jaegertracing/DynamicLoadTest.cpp
```diff
@@ -60,12 +61,14 @@ disabled: false
...
 #ifdef JAEGERTRACING_WITH_YAML_CPP
+
 TEST(DynamicLoad, invalidVersion)
 {
     const void* errorCategory = nullptr;
+    const void* errorMessage = nullptr;
     void* tracerFactory = nullptr;
-    const auto rcode = OpenTracingMakeTracerFactory(
-        "0.0.0" /*invalid version*/, &errorCategory, &tracerFactory);
+    const auto rcode = JaegerTracingMakeTracerFactoryFct(
```
isaachier: Why?

src/jaegertracing/Span.h
```diff
@@ -21,6 +21,8 @@
 #include <memory>
 #include <mutex>
 
+#include "jaegertracing/Compilers.h"
```
isaachier: `Config.h` should have included this everywhere already.

src/jaegertracing/TracerTest.cpp
```diff
@@ -250,6 +249,10 @@ TEST(Tracer, testTracer)
 
     span.reset();
 
+    // There is a problem here. After replacing the global tracer, the spans produced
+    // hold the last references to the tracer. After they are published, the tracer is
+    // destroyed. This is not the right moment to close the tracer.
+    tracer->Close();
```
isaachier: Why do you believe that is an issue? Who cares when the old tracer is closed?

src/jaegertracing/baggage/BaggageTest.cpp
```diff
@@ -105,7 +157,19 @@ TEST(Baggage, testRemoteRestrictionManagerFunctionality)
     ASSERT_EQ(Restriction(false, 0),
               manager.getRestriction("test-service", "abc"));
     mockAgent->addBaggageRestriction("abc", Restriction(true, 1));
-    std::this_thread::sleep_for(std::chrono::milliseconds(200));
+
+    //Wait for the Update Restrictions to occurr
```
isaachier: Typo on `occurr`.

src/jaegertracing/baggage/BaggageTest.cpp
```diff
@@ -105,7 +157,19 @@ TEST(Baggage, testRemoteRestrictionManagerFunctionality)
...
+    //This is done by listening to the number of update published via the Metrics
+    //Typically, On Windows with Debug mode, the update happens after 700ms
```
isaachier: Please add spaces before start of line comment.

src/jaegertracing/baggage/BaggageTest.cpp
```diff
@@ -105,7 +157,19 @@ TEST(Baggage, testRemoteRestrictionManagerFunctionality)
     ASSERT_EQ(Restriction(false, 0),
               manager.getRestriction("test-service", "abc"));
     mockAgent->addBaggageRestriction("abc", Restriction(true, 1));
-    std::this_thread::sleep_for(std::chrono::milliseconds(200));
+
+    //Wait for the Update Restrictions to occurr
+    //This is done by listening to the number of update published via the Metrics
```
isaachier: I think this brings in another component that is not needed. We can use another mechanism to wait for updates.

src/jaegertracing/net/Socket.h
```diff
@@ -169,22 +195,50 @@ class Socket {
...
             ::close(_handle);
+#endif
             _handle = -1;
         }
     }
 
-    int handle() { return _handle; }
+    HandleType handle() { return _handle; }
+
+    class OSResource {
```
isaachier: Please make this class `private`.

src/jaegertracing/platform/Endian.h
```diff
@@ -31,6 +31,31 @@
...
+#		define be16toh(x) (x)
+
+#		define htobe32(x) (x)
+#		define be32toh(x) (x)
+
+#		define htobe64(x) (x)
+#		define be64toh(x) (x)
+#	else
+#		error byte order not supported
+#   endif
```
isaachier: Please format the code changes using `scripts/clang-format.sh`.

src/jaegertracing/testutils/MockAgent.cpp
```diff
@@ -31,10 +31,36 @@
...
+        return std::max(0, readBytes); 
+#else
+        return ::read(socketHandle, buffer, size);
+#endif  
+    }
+
+    static size_t write(int socketHandle, const char* msg, size_t size)
+    {
+#ifdef WIN32
+        return ::send(socketHandle, msg, size, 0);
```
isaachier: `send` and `recv` work on Linux/Mac too. You can just use the Windows approach for all platforms and delete the original `read`/`write` calls.

src/jaegertracing/ThriftMethods.cpp
```diff
@@ -0,0 +1,71 @@
+/*
+ * Copyright (c) 2017 Uber Technologies, Inc.
```
isaachier: Copyright year is wrong.

src/jaegertracing/ThriftMethods.cpp
```diff
@@ -0,0 +1,71 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "jaegertracing/thrift-gen/BaggageRestrictionManager.h"
+#include "jaegertracing/thrift-gen/sampling_types.h"
+
+namespace jaegertracing {
+
+// trivial constructors of BaggageRestrictionManager thrift generated code are not generated by MSVC
```
isaachier: What?! Why not? This is why I am not so keen on supporting Windows.

CMakeLists.txt
```diff
@@ -27,6 +28,14 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unused-private-field")
 endif()
 
+if(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
```
ruslo: You can use `if(MSVC)` here.

CMakeLists.txt
ruslo: It depends on your goal/approach. You can use `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS` to export everything or use [GenerateExportHeader](https://cmake.org/cmake/help/latest/module/GenerateExportHeader.html) module and mark only particular classes/methods with `*_EXPORT`:
* https://github.com/cgold-examples/fruits/blob/081f76a970d3ebbc8d316ad8ef52994b8b4cad56/lib/fruits/rosaceae/Plum.hpp#L9
* https://github.com/cgold-examples/fruits/blob/081f76a970d3ebbc8d316ad8ef52994b8b4cad56/lib/fruits/rosaceae/Pear.hpp#L11

However I think if you decide to export everything then `set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)` should be set once in top level `CMakeLists.txt` and there is no need to unset it, `set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS OFF)` can be removed.

CMakeLists.txt
ruslo: I don't think it's related to Hunter, what is the exact reason of copying libraries?

isaachier: Well this project is on Hunter at the moment. I think there is some issue with Windows, but awaiting more detail from @mdouaihy.

CMakeLists.txt
isaachier: So I think I had wanted to do the export header thing, just never got around to that. I was just wondering on your opinion about `set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)`. It seems like something for `toolchain.cmake`. I know you do not believe `set(CMAKE_CXX_STANDARD 11)` is a good idea in CMakeLists.txt. I wonder if this could be a similar dangerous option.

CMakeLists.txt
mdouaihy: The problem I faced here is that the OT dynamic libraries were not accessible when running the Jaeger tests. On Linux, this works because of (or thanks to) the rpath.

CMakeLists.txt
mdouaihy: I had a dffierent approach at the beginning where I exported the symbols one by one. I found that fastidious and ended with a huge list. That's why I adopted the approach of exporting all symbols.

CMakeLists.txt
ruslo: > dynamic libraries were not accessible when running the Jaeger tests

You can modify PATH environment variable instead of copying libraries, something like that:
* https://github.com/cgold-examples/fruits/blob/081f76a970d3ebbc8d316ad8ef52994b8b4cad56/test/fruits/skin_off/CMakeLists.txt#L8-L27

src/jaegertracing/ThriftMethods.cpp
mdouaihy: I didnt dig a lot. For example, you can see that the copy ctor of BaggageRestrictionManager_getBaggageRestrictions_args is declared in the header but isnt defined.

isaachier:
@mdouaihy do you plan on updating this or are you busy?

mdouaihy:
Hello @isaachier, sorry for the delay to answer but I have been sick for the past weeks.

I will take your comments into consideration and update the PR soon.

isaachier:
Sorry to hear you were sick @mdouaihy. I based some of the code changes in #120 on some changes in dynamic loading you have here, so I appreciate the PR. Whenever you have a chance to update this, I'll be glad to help review it.

src/jaegertracing/TracerTest.cpp
mdouaihy: The problem is that the tracer is being destroyed. During that phase; the tracer will close the resources that are still used to report the the spans asynchronously.

src/jaegertracing/baggage/BaggageTest.cpp
mdouaihy: Hello @isaachier, what do you suggest to listen for the updates. Why do you think that listening for the metrics is a bad idea?

cmake/Hunter/config.cmake
mdouaihy: Hello @isaachier, HUNTER_BUILD_SHARED_LIBS impacts all the libraries imported by Hunter whereas the above line configure only OpenTracing as a shared Library. If I am not mistaken, this is what is suggested here https://github.com/ruslo/hunter/wiki/example.custom.config.id.

Why do you think is a bad idea?

@ruslo, what do you suggest? When I use HUNTER_BUILD_SHARED_LIBS, all libraries are considered as shared. The problem is that yaml-cpp does not build in such case.

src/jaegertracing/DynamicLoad.cpp
mdouaihy: I will revert this change.

src/jaegertracing/Span.h
mdouaihy: Why would I include Config.h in Span.h when I just need the compiler header fixing the MSVC compilation?

src/jaegertracing/baggage/BaggageTest.cpp
mdouaihy: Oh!!! Sorry for that. I will fix it.

CMakeLists.txt
isaachier: I'm thinking of going the route that jaeger-client-java did. Essentially, why export any of the symbols other than perhaps the config? The OpenTracing API is the only public API we recommend to users.

mdouaihy: I think that this can be a good idea. However, this means that won't be able to test the Jaeger code directly but we'll have to go through the OpenTracing API.

cmake/Hunter/config.cmake
ruslo: > When I use HUNTER_BUILD_SHARED_LIBS, all libraries are considered as shared. The problem is that yaml-cpp does not build in such case

Homogeneous approach (all static or all shared) is correct and easy to check, so `HUNTER_BUILD_SHARED_LIBS` should be used in most cases. If there are some troubles with `yaml-cpp` package then this package should be fixed.

isaachier: @ruslo I have one case where that is not ideal. This package is used as a plugin for https://github.com/envoyproxy/envoy. To do that, I need one big shared object with all dependencies packaged into it. I'll admit that it is pretty rare to use this pattern.

src/jaegertracing/DynamicLoad.cpp
isaachier: Thank you. This has been fixed on master. Try this `git rebase master`.

src/jaegertracing/Span.h
isaachier: I mean if `Compilers.h` is included in `Config.h`, then one of the includes below will bring `Compilers.h` here too. Please try to keep local imports at the end and sorted alphabetically. Don't rely on include ordering when you don't have to please.

src/jaegertracing/baggage/BaggageTest.cpp
isaachier: It isn't a bad approach, but if we have `InMemoryReporter` designed for this use case. Why not use that? I agree this test has race conditions, but I'd rather fix them instead of using another component.

mdouaihy: I dont think that the test has a race condition. it's just that some of the treatment is done asynchronously by another thread. We just need to wait to observe the result of that treatment. The reporter seemed to me as a way to observe such result.

CMakeLists.txt
mdouaihy: ok, I will do that.

cmake/Hunter/config.cmake
mdouaihy: One problem with YAML is that under MSVC, there is a problem with a macro definition. I issued a PR for that https://github.com/hunter-packages/yaml-cpp/pull/4

isaachier: Great. I am happy to see this was merged. We will need to prepare a new release of Hunter to land this review, but let's finish this part first.

CMakeLists.txt
mdouaihy: changed. With MSVC, a shared library is used. Otherwise, it'll be a static one.

CMakeLists.txt
mdouaihy: Changed. The path env variable is not set for the tests.

CMakeLists.txt
mdouaihy: change undone.

CMakeLists.txt
mdouaihy: change undone for Linux.

src/jaegertracing/DynamicLoad.cpp
mdouaihy: change reverted.

src/jaegertracing/ThriftMethods.cpp
mdouaihy: How can we proceed on this point? I think we can follow the bug in parallel, unless the thrift will dropped for protobuf.

mdouaihy:
@rnburn, @isaachier, @ruslo 
Sorry for the late PR update. I took me some time but I was able finally to allocate time for it.

CMakeLists.txt
```diff
@@ -203,12 +223,31 @@ function(add_lib_deps lib)
   target_link_libraries(${lib} PUBLIC ${LIBS})
 endfunction()
 
+function(update_path_for_test atest)
+  if(WIN32)
+  # If dependent library is a DLL we have to add location of DLL to PATH
```
isaachier: Can you please indent this block?

src/jaegertracing/Config.h
```diff
@@ -17,6 +17,8 @@
 #ifndef JAEGERTRACING_CONFIG_H
 #define JAEGERTRACING_CONFIG_H
 
+#include "jaegertracing/Compilers.h"
+
```
isaachier: Please remove newline here to keep with other Jaeger headers.

src/jaegertracing/LogRecord.h
```diff
@@ -17,6 +17,8 @@
 #ifndef JAEGERTRACING_LOGRECORD_H
 #define JAEGERTRACING_LOGRECORD_H
 
+#include "jaegertracing/Compilers.h"
```
isaachier: ~~As I've written before, this should be unnecessary. For instance, `Tag.h` should already include `Config.h`, which includes `Compilers.h`. Rule of thumb: you don't need to include a header already included in `Config.h`.~~

I'm sorry to say I got confused between this config header and one in the C client I have been working on (https://github.com/isaachier/jaeger-client-c/blob/master/src/jaegertracingc/config.h.in). My mistake. Keep this include wherever you need it.

src/jaegertracing/Span.h
```diff
@@ -21,6 +21,8 @@
 #include <memory>
 #include <mutex>
 
+//#include "jaegertracing/Compilers.h"
```
isaachier: Please remove this line.

src/jaegertracing/Reference.h
```diff
@@ -22,6 +22,8 @@
 
 #include <opentracing/propagation.h>
 
+#include "jaegertracing/Compilers.h"
+
```
isaachier: Newline issue.

src/jaegertracing/ThriftMethods.cpp
isaachier: There is no issue. You are correct. See if this works without the `#ifdef _MSVC`. It might not just be a Windows issue.

src/jaegertracing/Tracer.h
```diff
@@ -83,29 +84,7 @@ class Tracer : public opentracing::Tracer,
          const Config& config,
          const std::shared_ptr<logging::Logger>& logger,
          metrics::StatsFactory& statsFactory,
-         int options)
-    {
```
isaachier: Not a bad idea. I tried to inline a lot of code, but this is cleaner.

src/jaegertracing/TracerFactory.cpp
```diff
@@ -18,6 +18,13 @@
 
 #include "jaegertracing/Tracer.h"
 
+namespace YAML {
+namespace detail {
+
+//std::string node_data::empty_scalar;
+}
+}  // namespace YAML
```
isaachier: Please remove this block entirely.

src/jaegertracing/TracerTest.cpp
isaachier: If you see a crash, please report it as a separate issue. I would like to research this a little more. The real fix should never involve an explicit close. For now, please remove these fixes and we can rebase this on top of another PR for this issue.

src/jaegertracing/Transport.h
```diff
@@ -17,6 +17,8 @@
 #ifndef JAEGERTRACING_TRANSPORT_H
 #define JAEGERTRACING_TRANSPORT_H
 
+#include "jaegertracing/Compilers.h"
```
isaachier: Should appear last, where Jaeger headers always go. I believe `.clang-format` describes the ideal ordering.

src/jaegertracing/UDPTransport.h
```diff
@@ -17,6 +17,8 @@
 #ifndef JAEGERTRACING_UDPTRANSPORT_H
 #define JAEGERTRACING_UDPTRANSPORT_H
 
+#include "jaegertracing/Compilers.h"
+
```
isaachier: Please remove space.

src/jaegertracing/UDPTransportTest.cpp
```diff
@@ -100,7 +100,7 @@ TEST(UDPTransport, testExceptions)
         MockUDPClient::ExceptionType::kString
     };
     for (auto type : exceptionTypes) {
-        MockUDPTransport sender(net::IPAddress(), 0, type);
+        MockUDPTransport sender(net::IPAddress::v4("localhost", 0), 0, type);
```
isaachier: Why isn't this the default constructor?

src/jaegertracing/baggage/BaggageTest.cpp
```diff
@@ -21,12 +21,61 @@
...
+    }
+
+    void updateGauge(const std::string& name,
+                     int64_t time,
+                     const metrics::StatsReporter::TagMap& tags) override
+    {
+    }
+
+    int64_t counterValue(const std::string& name,
+                         const metrics::StatsReporter::TagMap& tags)
```
isaachier: `const` method?

src/jaegertracing/baggage/BaggageTest.cpp
```diff
@@ -105,7 +157,19 @@ TEST(Baggage, testRemoteRestrictionManagerFunctionality)
...
     mockAgent->addBaggageRestriction("abc", Restriction(true, 1));
-    std::this_thread::sleep_for(std::chrono::milliseconds(200));
+
+    // Wait for the Update Restrictions to occur
+    // This is done by listening to the number of update published via the Metrics
+    // Typically, On Windows with Debug mode, the update happens after 700ms
+    for (int sleepIndex = 0; sleepIndex < 10; ++sleepIndex) {
+        auto counterVal = testReporter.counterValue(
+            "jaeger.baggage-restrictions-update", { { "result", "ok" } });
+        std::this_thread::sleep_for(std::chrono::milliseconds(200));
```
isaachier: Probably don't need to sleep if `counterValue` waits on mutex.

src/jaegertracing/baggage/BaggageTest.cpp
```diff
@@ -21,12 +21,61 @@
...
+        std::lock_guard<std::mutex> lock(_mutex);
+
+        const auto metricName =
+            metrics::Metrics::addTagsToMetricName(name, tags);
+        return _counters[metricName];
+    }
+
+  private:
+    ValueMap _counters;
+    std::mutex _mutex;
```
isaachier: Use `volatile` keyword to allow for usage in `const` methods.

src/jaegertracing/baggage/RemoteRestrictionJSON.h
```diff
@@ -17,6 +17,8 @@
 #ifndef JAEGERTRACING_BAGGAGE_REMOTERESTRICTIONJSON_H
 #define JAEGERTRACING_BAGGAGE_REMOTERESTRICTIONJSON_H
 
+#include "jaegertracing/Compilers.h"
+
```
isaachier: Spacing...

src/jaegertracing/net/IPAddress.cpp
```diff
@@ -15,38 +15,92 @@
...
+
 namespace {
 
-struct IfAddrDeleter : public std::function<void(ifaddrs*)> {
-    void operator()(ifaddrs* ifAddr) const
+template <typename T>
+struct CDeleter : public std::function<void(T*)> {
+    void operator()(T* ifAddr) const
     {
         if (ifAddr) {
```
isaachier: I think both functions might work on a null argument. You can move `if` statement into `#else`.

src/jaegertracing/net/IPAddress.cpp
```diff
@@ -15,38 +15,92 @@
...
+        }
+    }
+
+    return IPAddress();
 }
 
-IPAddress IPAddress::localIP(std::function<bool(const ifaddrs*)> filter)
+#else
+
+#include <ifaddrs.h>
```
isaachier: I'd rather not include this mid-file. Can you please use `#ifndef WIN32` to wrap this include after the system includes at the top?

src/jaegertracing/net/IPAddress.cpp
```diff
@@ -58,6 +112,18 @@ IPAddress IPAddress::localIP(std::function<bool(const ifaddrs*)> filter)
     return IPAddress();
 }
 
+}
```
isaachier: Can you mark with `// anonymous namespace`?

src/jaegertracing/net/IPAddress.h
```diff
@@ -17,11 +17,7 @@
 #ifndef JAEGERTRACING_NET_IPADDRESS_H
 #define JAEGERTRACING_NET_IPADDRESS_H
 
-#include <arpa/inet.h>
-#include <netdb.h>
-#include <sys/socket.h>
-#include <sys/types.h>
-#include <unistd.h>
+#include "jaegertracing/Compilers.h"
```
isaachier: Please move to end of includes.

src/jaegertracing/net/Socket.h
```diff
@@ -169,22 +195,53 @@ class Socket {
...
+        OSResource& operator=(const OSResource&) = delete;
+        OSResource& operator=(const OSResource&&) = delete;
+    };
+
+    OSResource _osResource;
+    HandleType _handle;
     int _family;
     int _type;
+
+    bool isHandleInvalid(HandleType h)
```
isaachier: `static`?

src/jaegertracing/net/http/ResponseTest.cpp
```diff
@@ -24,14 +24,40 @@
...
 namespace jaegertracing {
 namespace net {
 namespace http {
 
+namespace {
+struct SocketTools {
+    static size_t read(int socketHandle, char* buffer, size_t size)
+    {
+#ifdef WIN32
+        int readBytes = ::recv(socketHandle, buffer, size, 0);
```
isaachier: Please use `recv`/`send` for non-Windows systems too.

src/jaegertracing/samplers/GuaranteedThroughputProbabilisticSampler.h
```diff
@@ -17,6 +17,8 @@
 #ifndef JAEGERTRACING_SAMPLERS_GUARANTEEDTHROUGHPUTPROBABILISTICSAMPLER_H
 #define JAEGERTRACING_SAMPLERS_GUARANTEEDTHROUGHPUTPROBABILISTICSAMPLER_H
 
+#include "jaegertracing/Compilers.h"
+
```
isaachier: Remove newline please.

src/jaegertracing/samplers/SamplerTest.cpp
```diff
@@ -39,7 +39,7 @@ constexpr auto kTestOperationName = "op";
 constexpr auto kTestFirstTimeOperationName = "firstTimeOp";
 constexpr auto kTestDefaultSamplingProbability = 0.5;
 constexpr auto kTestMaxID =
-    static_cast<long double>(std::numeric_limits<uint64_t>::max()) / 2.0 + 1;
+    std::numeric_limits<uint64_t>::max() / 2 + 1;
```
isaachier: @rnburn can explain the issue here (#6).

src/jaegertracing/testutils/MockAgent.cpp
```diff
@@ -31,10 +31,29 @@
...
+#ifdef _MSC_VER
+#pragma warning(disable : 4267)
+#pragma warning(disable : 4244)
+#endif
+
 namespace jaegertracing {
 namespace testutils {
 namespace {
 
+struct SocketUtils {
```
isaachier: What is this doing here again? And this one uses consistent behavior across platforms!

src/jaegertracing/utils/UDPClientTest.cpp
```diff
@@ -16,13 +16,13 @@
 
 #include "jaegertracing/net/IPAddress.h"
 #include "jaegertracing/net/Socket.h"
+
```
isaachier: Please remove newline. I think I used the include-what-you-use tool on the project, which reordered all the headers. I'd rather have consistency with style going forward.

src/jaegertracing/samplers/SamplerTest.cpp
mdouaihy: Unless I am missing something, in the issue #6, the problem occurs because we cast to double, which rounds up to a higher value. This latter, when casted back to  uint64_t cause the overflow.
Here, we are dividing an unsigned int by 2. I dont see why it can (over/under)flow.

src/jaegertracing/baggage/BaggageTest.cpp
mdouaihy: It's just a mutex and not a condition variable. We can have a spinning loop but we'll have to handle the timeout.

src/jaegertracing/testutils/MockAgent.cpp
mdouaihy: I will the remove those utility classes.
* I will replace write by ::send
* as for the read, I will keep two methods just to ajust the return value from signed to unsigned int.

src/jaegertracing/net/IPAddress.cpp
mdouaihy: Sorry @isaachier but I didnt get your point here.

src/jaegertracing/baggage/BaggageTest.cpp
isaachier: OK I understand the idea. If you'd like, you can add a condition variable too. Otherwise, this makes sense. Would prefer the condition variable to avoid sleeps, but will accept either solution.

src/jaegertracing/net/IPAddress.cpp
isaachier: Probably works:
```
void operator()(ifaddrs* ifAddr) const
{
#ifdef WIN32
    free(ifAddr);
#else
    ::freeifaddrs(ifAddr);
#endif
}
```

Definitely works:
```
#ifdef WIN32
    free(ifAddr);
#else
    if (ifAddr) {
        ::freeifaddrs(ifAddr);
    }
```
#endif
}

src/jaegertracing/testutils/MockAgent.cpp
isaachier: OK great. Just check for errors (negative numbers) before casting please.

src/jaegertracing/samplers/SamplerTest.cpp
isaachier: Indeed you are correct. My mistake.

src/jaegertracing/testutils/TUDPTransport.h
```diff
@@ -62,23 +68,34 @@ class TUDPTransport
...
     void write(const uint8_t* buf, uint32_t len)
     {
         ::sendto(_socket.handle(),
-                 buf,
-                 len,
-                 0,
-                 reinterpret_cast<const ::sockaddr*>(&_clientAddr.addr()),
-                 _clientAddr.addrLen());
+#ifdef WIN32
+            reinterpret_cast<const char*>(buf),  // this cast is safe
```
isaachier: That is fine with me. Any reason to make Windows specific and not on POSIX too?

src/jaegertracing/utils/UDPClient.h
```diff
@@ -59,7 +61,11 @@ class UDPClient : public agent::thrift::AgentIf {
                 << batch.spans.size();
             throw std::logic_error(oss.str());
         }
+#ifdef WIN32
+        const auto numWritten = ::send(_socket.handle(), reinterpret_cast<char*>(data), sizeof(uint8_t) * size, 0);
```
isaachier: Same question regarding casting only on Windows.

isaachier:
This is all great. I understand you probably had to redo your changes entirely for this second review. From now on, if possible, please commit on top of the older changes and do not squash so we can see updates more easily.

src/jaegertracing/testutils/TUDPTransport.h
mdouaihy: On POSIX, send/sendto expect void*. It will be weird to cast to char*. What do you think?

src/jaegertracing/testutils/MockAgent.cpp
mdouaihy: in fact, I used std::max(0, readBytesSize). This allowed t return 0 in case of error. Is that ok for you?

src/jaegertracing/testutils/TUDPTransport.h
isaachier: Even if weird better than macros IMO.

src/jaegertracing/testutils/MockAgent.cpp
isaachier: I don't understand. How does windows indicate a read was unsuccessful as opposed to successful but zero bytes read.

src/jaegertracing/UDPTransportTest.cpp
mdouaihy: because the behavior on windows is a bit different.
In `UDPClient::UDPClient`, `v_socket.connect(_serverAddr);` throws an exception: 
**Cannot connect socket to remote address { family=2, addr=0.0.0.0, port=0 }**

src/jaegertracing/testutils/MockAgent.cpp
mdouaihy: Here is the doc for both
* Linux
  * http://man7.org/linux/man-pages/man2/recvmsg.2.html
  * recv return an unsigned int
  * In case of error, it returns -1 (that will underflow to max int)
* Windows 
  * https://docs.microsoft.com/en-us/windows/desktop/api/winsock/nf-winsock-recv
  * recv return an int
  * In case of error, it returns -1

Now that I am answering you, I find myself thinking that I should test again the error value and return 0 or the value.

src/jaegertracing/TracerTest.cpp
mdouaihy: Ok I see. In fact, the point was not fix the crash but to avoid having the crash until it's fixed.

I will deactivate this test for the moment under windows and report it as an issue with the details.

isaachier: Sure, but I want to try fixing that bug first. If it takes more than a day or two, we can disable the test to land this PR.

src/jaegertracing/testutils/MockAgent.cpp
mdouaihy: Here is the doc for both
* Linux
  * http://man7.org/linux/man-pages/man2/recvmsg.2.html
  * recv return an unsigned int
  * In case of error, it returns -1 (that will underflow to max int)
* Windows 
  * https://docs.microsoft.com/en-us/windows/desktop/api/winsock/nf-winsock-recv
  * recv return an int
  * In case of error, it returns -1

Now that I am answering you, I find myself thinking that I should test again the error value and return 0 or the value.

isaachier: Looking through the original code, I see we handled the case of negative values correctly. My suggestion is just to replace `read`/`write` with `recv`/`send` where feasible. The return values will be handled correctly.

isaachier: For the record, `ssize_t` is a *signed* size type, so there will be no underflow when it returns -1.

src/jaegertracing/TracerTest.cpp
mdouaihy: Issue reported #127

FredrikAppelros:
Is work on this progressing? Do you need any help? Would love to see this merged ASAP.

mdouaihy:
Hi @FredrikAppelros, some tests are failing on Windows because if the bug described in issue #127.

I believe this is the last issue to solve in order to have this PR pushed.

isaachier:
> Hi @FredrikAppelros, some tests are failing on Windows because if the bug described in issue #127.
> 
> I believe this is the last issue to solve in order to have this PR pushed.

Hey I'll try to look this over tonight. If that is the only issue, I will allow a quick hack seeing as I have not come up with an easy fix for the bug.

solomonlu:
Hello everyone, can this PR continue? We are eager to use jaeger on Windows.

isaachier:
I'll try to see if this still merges and looks good.

isaachier:
I see I have a bunch of unresolved comments here. Please fix them and resubmit.

FredrikAppelros:
@isaachier: Are you referring to the style changes you requested or something more substantial when you mention unresolved comments?

As I offered before, I can help if I were given permission to push commits to this PR from @mdouaihy. The style changes should be easy enough, and I also have some fixes on top of this locally that I would like to see here.

One of the issues I had prevented me from getting this to compile since it could not pull in some dependency properly and the other makes sure that we can target the Windows 8.1 SDK.

I have not looked into the failing test case you guys mention though. Would be great if you could work out a solution for that.

isaachier:
@FredrikAppelros I have no control over the PR source, but you can always fork the windowsbuild branch from @mdouaihy and make the changes there. If you do that, we can open a PR based off of your version of windowsbuild and potentially land this without @mdouaihy's involvement. If @mdouaihy is free to help, then there is a better way to do this: make a PR onto @mdouaihy's windowsbuild branch and it will automatically update here when it lands there.

yurishkuro:
Or make smaller incremental PRs.

mdouaihy:
hello guys, I will recheck the changes requested by @isaachier. However, i still have the problem described in issue https://github.com/jaegertracing/jaeger-client-cpp/issues/127. That blocked me from going further.

isaachier:
@mdouaihy I was/am ready to accept this PR despite #127, especially because there is no easy fix there. You can use the local fix you had previously used. Just try to fix other comments and style suggestions.

efficks:
Hi, Is anybody have any news on this PR. Is someone needs help to make it merged? I offer my help if needed.

isaachier:
> Hi, Is anybody have any news on this PR. Is someone needs help to make it merged? I offer my help if needed.

Hey @efficks thanks for the offer. Last I heard, @mdouaihy was going to look over the changes. If you'd like to take over this PR, try pulling @mdouaihy's branch and writing a PR against @mdouaihy's fork. Then, if @mdouaihy accepts it on this branch, it will automatically update the PR.

mdouaihy:
Hello,
I am working again on the PR. However, due to update of msvc to 15.9, some dependencies dont compile anymore. I am following/working on them.

cmake/Hunter/config.cmake
mdouaihy: I did a new fix in Yaml-cpp this week. Hunter updated https://github.com/ruslo/hunter/releases/tag/v0.23.202

src/jaegertracing/Compilers.h
mdouaihy: Shall I update to 2019?

src/jaegertracing/net/Socket.h
mdouaihy: done

src/jaegertracing/testutils/MockAgent.cpp
mdouaihy: done

src/jaegertracing/ThriftMethods.cpp
mdouaihy: 2019?

CMakeLists.txt
mdouaihy: done

CMakeLists.txt
mdouaihy: done

src/jaegertracing/Config.h
mdouaihy: done

src/jaegertracing/Span.h
mdouaihy: done

src/jaegertracing/testutils/TUDPTransport.h
mdouaihy: I personally prefer the avoiding the cast when not needed.

mdouaihy:
Hello @isaachier, I updated the PR with the changes you requested and rebased the version of the PR.

Would you please check and review it?

Sorry for the delay, but I had some personal stuff that prevented me from checking this PR before.

FredrikAppelros:
Any updates regarding this?

FredrikAppelros:
Sorry for nagging, just wondering if it is possible to get this reviewed?

mdouaihy:
Hello @isaachier, any news about the review?

CMakeLists.txt
```diff
@@ -210,12 +230,31 @@ function(add_lib_deps lib)
...
+  endif()
+endfunction()
+
 option(JAEGERTRACING_PLUGIN "Build dynamic plugin" OFF)
 cmake_dependent_option(BUILD_SHARED_LIBS "Build shared libraries" ON
   "NOT JAEGERTRACING_PLUGIN" OFF)
 
 if(BUILD_SHARED_LIBS)
   add_library(jaegertracing SHARED ${SRC})
+if (WIN32)
```
yurishkuro: should this be indented more?

src/jaegertracing/Tracer.cpp
```diff
@@ -214,4 +214,32 @@ Tracer::analyzeReferences(const std::vector<OpenTracingRef>& references) const
     return result;
 }
 
+std::shared_ptr<opentracing::Tracer>
+Tracer::make(const std::string& serviceName,
```
yurishkuro: Is this just moving this func from .h ?

src/jaegertracing/UDPTransport.cpp
```diff
@@ -27,6 +27,10 @@
 #include <thrift/protocol/TCompactProtocol.h>
 #include <thrift/transport/TBufferTransports.h>
 
+#ifdef _MSC_VER
+#pragma warning(disable : 4267)
```
yurishkuro: it would be useful to add comments about why something is being disabled

yurishkuro:
@isaachier hasn't been working on Jaeger for some time now, so he may not be around to review this PR. 

Is there a way to test that the Windows build is actually fixed? I see that `.appveyor.yml` is added, but there's no CI status posted from appveyor. Has the project been set up there?

CMakeLists.txt
mdouaihy: indeed.

src/jaegertracing/Tracer.cpp
mdouaihy: yes.

src/jaegertracing/baggage/BaggageTest.cpp
mdouaihy: mutable used.

src/jaegertracing/Compilers.h
mdouaihy: udated to 2019

src/jaegertracing/utils/UDPClient.h
mdouaihy: changed to ::send for Linux and Windows.

src/jaegertracing/UDPTransport.cpp
mdouaihy: ok

mdouaihy:
Hi @yurishkuro, I configured a build for my fork at https://ci.appveyor.com/project/mdouaihy/jaeger-client-cpp/builds with appveyor configuration. I am thinking of keeping only Debug or Release instead of having them both (leading to doubled build time). What do you think?

yurishkuro:
@mdouaihy what do you mean by "keeping debug or release only"? keeping where?

we seem to have an official appvoyer account for this repo, but it has no builds since last year.

https://ci.appveyor.com/project/jaegertracing/jaeger-client-cpp/settings

Would you like to configure it to build for this PR? I sent you an invitation to that project.

mdouaihy:
Hi @yurishkuro, my question was about the types of builds to run under windows: Debug and/or Release. I ended up keeping only the Debug build since the release one does offer any added value.

As for the integration with Appveyor, I dont have permission on the github project. Can you check that the Github Appveyor app is well configured to run on PR?

yurishkuro:
I reinstalled AppVeyor GitHubApp.

I see this in the Events
image

But in the build history I only see older builds.

mdouaihy:
I think that this is because it's configured to run only on the master and not for the PR, no?

yurishkuro:
I don't think so.
image

image

mdouaihy:
I pushed a small change to the PR. Appveyor was triggered.

src/jaegertracing/TracerTest.cpp
```diff
@@ -218,14 +219,15 @@ TEST(Tracer, testTracer)
             ->StartSpanWithOptions("test-span-with-default-system-timestamp",
                                    options)
             .release()));
+    /*
```
yurishkuro: please add a comment, or better a link to an issue describing why this test is being disabled

src/jaegertracing/TracerTest.cpp
yurishkuro: similarly, add link to the issue in the comments under TODO:

src/jaegertracing/TracerTest.cpp
mdouaihy: ok. Issue inserted #167 and issue referenced in the code.

mdouaihy:
it seems that Appveyor does both: update the check and submit a message.

yurishkuro:
Hm, I don't get it, now the Travis status is gone. I was hoping to get both, testing on different OSes

mdouaihy:
hi @yurishkuro, maybe we need to reinstall travis ci?

mdouaihy:
hi @yurishkuro, any update?

yurishkuro:
I disabled and re-enabled the repo in Travis, and restarted the last build (which was for commit  d08082c, so it missed several latest commits).

yurishkuro:
Can you push another commit to this PR?

mdouaihy:
Thank you @yurishkuro !

yurishkuro:
thank YOU for your contribution.

FredrikAppelros:
Excellent! Will we see this in a release soon? :)

yurishkuro:
According to https://github.com/jaegertracing/jaeger-client-cpp/blob/master/RELEASE.md, releasing is just a tag?  If someone can create the pre-release PR (changelog changes, etc), then I can do a release.

mdouaihy:
PR #169 sent for that purpose.

Issue #118, opened 2018-07-30
update hunter version to use a newer opentracing-cpp?


ppggff:
The build-in version hunter will build with opentracing-cpp 1.4.0, and the "plugin so" will return error with other version.

isaachier:
Good catch. Will update now. Thanks.

isaachier:
@rnburn, while doing this, discovered that Jaeger does not implement new OpenTracing dynamic loading function. Where do I use `OPENTRACING_DECLARE_IMPL_FACTORY`? In the implementation file (DynamicLoad.cpp) or in the tracer factory header (TracerFactory.h)?

ppggff:
It would be better to add the version number of corresponding opentracing-cpp with the released binary plugin library file.

isaachier:
Fixed in #120.

PR #120, opened 2018-07-31
Upgrade dynamic loading API

isaachier:
- #118 

- Update Hunter and upgrade dynamic loading API.


src/jaegertracing/DynamicLoad.cpp
```diff
@@ -19,31 +19,37 @@
...
 #ifndef JAEGERTRACING_WITH_YAML_CPP
     *errorCategory =
         static_cast<const void*>(&opentracing::dynamic_load_error_category());
     return opentracing::dynamic_load_not_supported_error.value();
 #endif
-    if (std::strcmp(opentracingVersion, OPENTRACING_VERSION) != 0) {
-        *errorCategory = static_cast<const void*>(
+    if (std::strcmp(opentracing_version, OPENTRACING_VERSION) != 0 ||
+        std::strcmp(opentracing_abi_version, OPENTRACING_ABI_VERSION) != 0) {
+        *error_category = static_cast<const void*>(
```
rnburn: I don't think you want `std::strcmp(opentracing_version, OPENTRACING_VERSION) != 0` in the comparison.

The point of adding the ABI version was to allow the plugin to work with other versions of OT if there's no ABI breakage.

The version is only passed in for the case that you might have some sort of incompatible semantic change in the future.

src/jaegertracing/DynamicLoad.h
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+#ifndef JAEGERTRACING_DYNAMIC_LOAD_H
+#define JAEGERTRACING_DYNAMIC_LOAD_H
+
+int JaegerTracingMakeTracerFactoryFct(const char* opentracing_version,
+                                      const char* opentracing_abi_version,
+                                      const void** error_category,
+                                      void* error_message,
+                                      void** tracer_factory);
```
rnburn: You don't need the function exposes. The macro OPENTRACING_DECLARE_IMPL_FACTORY declares a variable function pointer (See for example how it's done in the LS tracer: https://github.com/lightstep/lightstep-tracer-cpp/blob/master/src/dynamic_load.cpp#L6)

src/jaegertracing/DynamicLoad.cpp
isaachier: Good point.

src/jaegertracing/DynamicLoad.h
isaachier: Thanks for the example!

isaachier:
@rnburn do you mind checking if everything looks OK? Thanks in advance.

src/jaegertracing/DynamicLoad.cpp
```diff
@@ -19,21 +19,23 @@
...
+                             const void** errorCategory,
+                             void* errorMessage,
+                             void** tracerFactory)
 {
     assert(errorCategory != nullptr);
     assert(tracerFactory != nullptr);
 #ifndef JAEGERTRACING_WITH_YAML_CPP
     *errorCategory =
-        static_cast<const void*>(&opentracing::dynamic_load_error_category());
+        static_cast<const void*>(&opentracing::dynamic_load_errorCategory());
```
rnburn: dynamic_load_errorCategory?

isaachier: Lol I was camelcasing this stuff an must have messed that up. Surprised this built...

PR #130, opened 2018-09-13
localIP: try resolving local hostname first

### LINKED ISSUE #129: The C++ Jaeger client always populates the "ip" tag with 127.0.0.1

The C++ Jaeger client always populates the "ip" tag with 127.0.0.1 in my environment (Docker containers running on a local machine), which makes the tag useless.

Incorrect implementation in https://github.com/jaegertracing/jaeger-client-cpp/blob/7e9a135b362c47f7e4d42e0693b05b91a70e6888/src/jaegertracing/net/IPAddress.cpp#L45: Looking at the implementation, this will return the address of the first encountered `AF_INET` interface, which could be the loopback interface.

I know that getting a node's IP address is an ambiguous question with no right answer -- which interface do you use? Many nodes have more than one IP address, especially in the network-virtualized environments of Kube clusters, but the loopback interface is almost definitely the *wrong* answer.

A better solution would allow overriding the IP address via environment variables and/or the tracer config, as per #105.

tudor:
Fixes #129.

Signed-off-by: Tudor Bosman <tudor@rockset.com>

- #129 

- Just like the Jaeger Java client does, resolve the local hostname in order to determine the local IP address (to populate the "ip" tracer tag with).

### DIFF
--- a/src/jaegertracing/net/IPAddress.cpp
+++ b/src/jaegertracing/net/IPAddress.cpp
@@ -15,6 +15,7 @@
  */
 
 #include "jaegertracing/net/IPAddress.h"
+#include "jaegertracing/platform/Hostname.h"
 
 #include <ifaddrs.h>
 #include <sys/types.h>
@@ -36,6 +37,12 @@ struct IfAddrDeleter : public std::function<void(ifaddrs*)> {
 
 IPAddress IPAddress::localIP(int family)
 {
+    try {
+        return versionFromString(platform::hostname(), 0, family);
+    } catch (const std::exception& e) {
+        // Fall back to returning the first matching interface
+    }
+
     return localIP([family](const ifaddrs* ifAddr) {
         return ifAddr->ifa_addr != nullptr &&
                ifAddr->ifa_addr->sa_family == family;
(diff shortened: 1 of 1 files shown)


tudor:
The test failure seems unrelated:

```
[ RUN      ]Reporter.testRemoteReporter
/home/travis/build/jaegertracing/jaeger-client-cpp/src/jaegertracing/reporters/ReporterTest.cpp:80: Failure
Expected equality of these values:
  i + 1
    Which is: 72
  spans.size()
    Which is: 71
[  FAILED  ]Reporter.testRemoteReporter (333 ms)
```

src/jaegertracing/net/IPAddress.cpp
```diff
@@ -36,6 +37,12 @@ struct IfAddrDeleter : public std::function<void(ifaddrs*)> {
 
 IPAddress IPAddress::localIP(int family)
 {
+    try {
+        return versionFromString(platform::hostname(), 0, family);
+    } catch (const std::exception& e) {
```
isaachier: If you don't care about the exception, why not use `catch (,..)` to catch everything?

tudor:
Updated to `catch (...)`. PTAL?

Issue #131, opened 2018-09-15
Jaeger client deadlocks with span of duration 0


tudor:
## Requirement - what kind of business use case are you trying to solve?

I have a deadlock in a process that uses the Jaeger C++ client. I debugged it to show a deadlock in the case where a span's duration is zero. Details in the PR.

## Problem - what in Jaeger blocks you from solving the requirement?

Jaeger bug in `Span::FinishWithOptions`.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Sending a PR.

isaachier:
Fixed in #132.

Issue #127, opened 2018-09-06
Span shared_ptr to Tracer causes deadlock in RemoteReporter


mdouaihy:
## Problem 
I am experiencing random failures of the test `Tracer, testTracer`.

After investigation, here is what I found:
- `Tracer, testTracer` starts a Reporter in a separate thread. Note that on a busy machine, this thread might be slow to consume the reported span
- Sometimes, the test finishes quickly before the reported had a chance to report the span
- In this case
  - the test replace the tracer with the Noop one at the end. At this point, the already reported spans become the only references to the previous tracer
  - When the reporter finally has a chance to run, it sends the spans and destroys them
  - Once the last span is destroyed, the tracer gets destroyed also. Among other resources , the tracer tries to destroy the reporter that is being running. 

I observed here two different behaviors

- Under Windows, the mutex are recursives. The attempt to destroy the reporter succeed. The sweep method then reads dangling data
- Under macOS, Linux, the mutex is not recursive. When calling ``RemoteReporter::close()`` blocks when called from ``RemoteReporter::sweepQueue``. The reporter thread indefinitely.

For additional analysis, below are the stack traces under Windows and macOS.

**Under Windows**. [External Code] is the standard library for which I didnt load the symbols.

``
>	jaegertracingd.dll!`jaegertracing::utils::ErrorUtil::logError'::`1'::catch$7() Line 41	C++
 	[External Code]	
 	jaegertracingd.dll!jaegertracing::utils::ErrorUtil::logError(jaegertracing::logging::Logger &, const std::basic_string<char,std::char_traits<char>,std::allocator<char> > & {...}) Line 33	C++
 	jaegertracingd.dll!`jaegertracing::reporters::RemoteReporter::close'::`1'::catch$2() Line 77	C++
 	[External Code]	
 	jaegertracingd.dll!jaegertracing::reporters::RemoteReporter::close() Line 67	C++
 	jaegertracingd.dll!jaegertracing::reporters::RemoteReporter::~RemoteReporter() Line 47	C++
 	[External Code]	
 	jaegertracingd.dll!jaegertracing::Tracer::~Tracer() Line 89	C++
 	[External Code]	
 	jaegertracingd.dll!jaegertracing::Span::~Span() Line 84	C++
 	jaegertracingd.dll!jaegertracing::reporters::RemoteReporter::sweepQueue() Line 100	C++
[...]

PR #132, opened 2018-09-15
Fix deadlock if steady_clock::now() returns the same value twice

tudor:
`Span::isFinished()` returns true iff the span's duration is zero. So, if `steady_clock::now()` returns the same value at span creation as at `Finish()` (unlikely, but possible, especially in virtualized environments), the following scenario can happen:

- `Span::FinishWithOptions` sets `_duration` to zero
- and then calls `reportSpan`, which calls `RemoteReporter::report`
- which acquires the reporter mutex and
- adds the span to the reporter's `_queue`

Later:
- `RemoteReporter::sweepQueue` acquires the reporter mutex
- `RemoteReporter::sweepQueue` makes a copy of the span:
https://github.com/jaegertracing/jaeger-client-cpp/blob/7e9a135b362c47f7e4d42e0693b05b91a70e6888/src/jaegertracing/reporters/RemoteReporter.cpp#L95
- The span is serialized in `RemoteReporter::sendSpan`
- The span (the copy made above) needs to be destroyed because the block in `sweepQueue` ends.
- The span's destructor is called, which calls `FinishWithOptions` again
- ... which sees that `_duration` is zero, so `isFinished()` returns false:
https://github.com/jaegertracing/jaeger-client-cpp/blob/7e9a135b362c47f7e4d42e0693b05b91a70e6888/src/jaegertracing/Span.cpp#L94
- so it tries to call `reportSpan` again
- which tries to acquire the reporter's mutex
- but we're holding the lock in `sweepQueue`, so
- deadlock.

Relevant stack trace:
```
(gdb) bt
 #0  __lll_lock_wait () at ../sysdeps/unix/sysv/linux/x86_64/lowlevellock.S:135
 #1  0x00007fb09c722023 in __GI___pthread_mutex_lock (mutex=0x7fb088b1c1c0) at ../nptl/pthread_mutex_lock.c:78
 #2  0x00007fb08860987b in jaegertracing::reporters::RemoteReporter::report(jaegertracing::Span const&) () from /usr/local/lib/libjaegertracing_plugin.so
 #3  0x00007fb0885b6506 in jaegertracing::Span::FinishWithOptions(opentracing::v2::FinishSpanOptions const&) [clone .localalias.361] () from /usr/local/lib/libjaegertracing_plugin.so
 #4  0x00007fb08860849a in jaegertracing::Span::~Span() [clone .constprop.253] () from /usr/local/lib/libjaegertracing_plugin.so
 #5  0x00007fb08860a8fa in jaegertracing::reporters::RemoteReporter::sweepQueue() () from /usr/local/lib/libjaegertracing_plugin.so
 #6  0x00007fb0886f2f5f in execute_native_thread_routine () from /usr/local/lib/libjaegertracing_plugin.so
 #7  0x00007fb09c71f6db in start_thread (arg=0x7fb087574700) at pthread_create.c:463
 #8  0x00007fb0989ae88f in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```

Fixed by enforcing a minimum span duration of 1 `SteadyClock` tick.

Signed-off-by: Tudor Bosman <tudor@rockset.com>

- #131 

- Fix deadlock when span duration is zero by enforcing a minimum span duration of 1 tick

### DIFF
--- a/src/jaegertracing/Span.cpp
+++ b/src/jaegertracing/Span.cpp
@@ -96,6 +96,12 @@ void Span::FinishWithOptions(
             return;
         }
         _duration = finishTimeSteady - _startTimeSteady;
+        if (_duration <= SteadyClock::duration()) {
+            // Enfoce minimum duration of 1 tick (1ns on Linux),
+            // so isFinished() returns true
+            _duration = SteadyClock::duration(1);
+        }
+
         tracer = _tracer;
 
         std::copy(finishSpanOptions.log_records.begin(),
--- a/src/jaegertracing/Span.cpp
+++ b/src/jaegertracing/Span.cpp
@@ -97,7 +97,7 @@ void Span::FinishWithOptions(
         }
         _duration = finishTimeSteady - _startTimeSteady;
         if (_duration <= SteadyClock::duration()) {
-            // Enfoce minimum duration of 1 tick (1ns on Linux),
+            // Enforce minimum duration of 1 tick (1ns on Linux),
             // so isFinished() returns true
             _duration = SteadyClock::duration(1);
         }


tudor:
CI fails in the same way as https://github.com/jaegertracing/jaeger-client-cpp/pull/130. That test is flaky.

```
[ RUN      ]Reporter.testRemoteReporter
/home/travis/build/jaegertracing/jaeger-client-cpp/src/jaegertracing/reporters/ReporterTest.cpp:80: Failure
Expected equality of these values:
  i + 1
    Which is: 59
  spans.size()
    Which is: 58
[  FAILED  ]Reporter.testRemoteReporter (277 ms)
```

isaachier:
CI failure is a real issue I need to fix. Just added #134. The deadlock actually seems to be related to #127. See issue there and let me know if you still think this is the answer.

tudor:
> The deadlock actually seems to be related to #127. See issue there and let me know if you still think this is the answer.

#127 has to do with destroying the `Tracer` object (see `Tracer::~Tracer` on the stacks there), which is not happening here. In our case (in production), the tracer is global, created at process startup, and never deleted until the process exits.

These are two separate issues.

src/jaegertracing/Span.cpp
```diff
@@ -96,6 +96,12 @@ void Span::FinishWithOptions(
             return;
         }
         _duration = finishTimeSteady - _startTimeSteady;
+        if (_duration <= SteadyClock::duration()) {
+            // Enfoce minimum duration of 1 tick (1ns on Linux),
```
isaachier: Sorry for nit, but typo `Enforce`.

tudor:
Fixed the typo. PTAL?

tudor:
Thanks! Are you planning to cut a new release any time soon? (I'm okay using the master branch as well, just curious.)

isaachier:
I won't stop you if you'd like to release a minor version. See RELEASE.md for how to do it. I can build the plugin or do the entire release, but I'd like to see others take a stab at it.

PR #138, opened 2018-11-21
Implement SpanContext::Clone

meastman:
This was method added to the opentracing API as a pure virtual method in <https://github.com/opentracing/opentracing-cpp/pull/56>. Without implementing this method, the jaeger client library cannot be used with recent (post-August-2018) versions of opentracing.

### DIFF
--- a/src/jaegertracing/SpanContext.h
+++ b/src/jaegertracing/SpanContext.h
@@ -19,6 +19,7 @@
 
 #include <iomanip>
 #include <iostream>
+#include <memory>
 #include <mutex>
 #include <string>
 #include <unordered_map>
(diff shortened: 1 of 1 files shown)


isaachier:
@meastman looks good. Travis CI has this error:

```
In file included from /home/travis/build/jaegertracing/jaeger-client-cpp/src/jaegertracing/SpanContext.cpp:17:0:
/home/travis/build/jaegertracing/jaeger-client-cpp/src/jaegertracing/SpanContext.h:167:10: error: ‘unique_ptr’ in namespace ‘std’ does not name a template type
     std::unique_ptr<opentracing::SpanContext> Clone() const noexcept
```

Try including `<memory>` in `SpanContext.h`.

src/jaegertracing/SpanContext.h
```diff
@@ -164,6 +164,12 @@ class SpanContext : public opentracing::SpanContext {
         forEachBaggageItem(f);
     }
 
+    std::unique_ptr<opentracing::SpanContext> Clone() const noexcept
+    {
```
isaachier: Can you add a lock guard here please:

`std::lock_guard<std::mutex> lock(_mutex);`

meastman:
> LGTM thanks for the PR.

Thanks for the quick review!

PR #145, opened 2018-12-26
Change sampler param sentinel value from YAML parser

### LINKED ISSUE #144: Getting sampled spans when using const sampler and param 0

I am trying to disable all non-debug sampling, so that I specifically get traces only when I ask for them via the debug header.

I am using the following config:
```json
{
  "service_name": "avm-nginx",
  "sampler": {
        "type": "const",
        "param": 0
  },
  "reporter": {
        "localAgentHostPort": "jaeger-agent.tracing:6831"
  }
}
```
via the nginx-opentracing module, but traces are still being sampled.

Make const sampler work as documented when param is set to 0.

I believe the issue is that if the param option is not set, the yaml config parser sets it to 0. A value of 0 is later interpreted as the param being not set, so the param is set to 0.001 because that is default probabilistic sampling rate. Then even later, any non-zero param is interpreted as always sample for the ConstSampler, so there is no way to make the ConstSampler have a false decision. I don't expect this to be intended behavior.

If some other sentinel value was used, like -1, then specifying param = 0 would make ConstSampler and ProbablisticSampler work as documented, and not specifying the param would still work the same.

andremarianiello:
- Resolves #144

- Instead of using 0 as the sentinel value for an unspecified sampler param, use another number that would be illegal instead. This allows param = 0 to be specified and respected by samplers.

### DIFF
--- a/src/jaegertracing/ConfigTest.cpp
+++ b/src/jaegertracing/ConfigTest.cpp
@@ -73,6 +73,18 @@ TEST(Config, testDefaultSamplingProbability)
               Config().sampler().param());
 }
 
+TEST(Config, testZeroSamplingParam)
+{
+    {
+        constexpr auto kConfigYAML = R"cfg(
+sampler:
+    param: 0
+)cfg";
+        const auto config = Config::parse(YAML::Load(kConfigYAML));
+        ASSERT_EQ(0, config.sampler().param());
+    }
+}
+
 #endif  // JAEGERTRACING_WITH_YAML_CPP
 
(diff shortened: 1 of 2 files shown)


yurishkuro:
is it possible to add a unit test?

andremarianiello:
@yurishkuro Added a test, which succeeds, and I verified that it failed without the sentinel value changes.

isaachier:
Let me take a look soon.

andremarianiello:
Is there anything else I need to do for this?

PR #158, opened 2019-05-23
Update default samplingServerURL to include /sampling path

### LINKED ISSUE #157: Default sampling server url misses /sampling path
At https://github.com/jaegertracing/jaeger-client-cpp/blob/v0.4.2/src/jaegertracing/samplers/Config.h#L45 the default sampling server url for retrieving remote sampling strategies is set to `http://127.0.0.1:5778`, but this should be `http://127.0.0.1:5778/sampling` instead.

If you retrieve `http://127.0.0.1:5778/?service=my-service` the jaeger-agent returns

```json
{"strategyType":0,"probabilisticSampling":{"samplingRate":0.5}}
```

While at the correct url `http://127.0.0.1:5778/sampling?service=my-service` it returns

```json
{"strategyType":"PROBABILISTIC","probabilisticSampling":{"samplingRate":0.5}}
```

If the sampling strategies file looks like this:

```json
{
  "service_strategies": [
  ],
  "default_strategy": {
    "type": "probabilistic",
    "param": 0.5
  }
}
```

JorritSalverda:
Resolves #157

Signed-off-by: Jorrit Salverda <jorrit.salverda@gmail.com>

The default sampling server url misses the /sampling path, which leads to an incorrect response and inability to use remote configuration.

Updated the default url and added a test.


JorritSalverda:
I have no experience writing C++, so let me know if I do something unusual for the language. Although there's hardly any coding involved :)

yurishkuro:
Have you tested it against Jaeger server?  I don't think the `crossdock` integration test in this repo actually tests polling for sampling strategies, it only tests spans submission. I believe there  is a difference between `/` and `/sampling` endpoints in the returned data format.

hiruna:
hi can we get this merged ?

yurishkuro:
I would like to see some proof of testing with the agent.

hiruna:
@yurishkuro , there is a difference between `/` and `/sampling` (correct me if im wrong), `/` returns the default value for a `remote` sampler, which is `probabilistic | 0.5` this was mentioned in https://github.com/jaegertracing/jaeger-client-cpp/issues/157

yurishkuro:
The difference is not in the semantics of the response but in the format of the data, how the thrift enums were encoded into json (numbers instead of strings).

skwiffhub:
@yurishkuro Yes, thanks, we were able to test with the fix. We did not verify the [`ConfigTest.cpp` portion of the commit](https://github.com/jaegertracing/jaeger-client-cpp/pull/158/commits/63928aaab94149cb57514cb731afdb5f0ab7cf89) though.

One other thing to check for is sample instances of `config.yml` files that contain `samplingServerURL` and make sure they are updated to reflect what is in the current source code.

yurishkuro:
@skwiffhub thanks for the confirmation, I will merge it. I am not too concerned with config test because it actually has a unit test, but there's no integration test with the real agent.

yurishkuro:
Thanks for the PR, @JorritSalverda

hiruna:
Is this going to be released soon ?

hiruna:
any update on this @yurishkuro @JorritSalverda

PR #169, opened 2019-09-05
Preparing release 0.5.0

mdouaihy:
Signed-off-by: Mehrez DOUAIHY <mehrez.douaihy@gmail.com>

- Prepare for release 0.4.3

- Prepare the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,18 @@
 Changes by Version
 ==================
 
-0.5.0 (unreleased)
+0.5.0 (2019-09-07)
 ------------------
-- Nothing yet
+- Improve C++ flag handling in CMake (#128)
+- localIP: try resolving local hostname first (#130)
+- Fix deadlock if steady_clock::now() returns the same value twice (#132)
+- Implement SpanContext::Clone (#138)
+- Change sampler param sentinel value from YAML parser (#145)
+- Fix RemoteReporter test race condition (#135)
+- Add details on how to set agent & collector endpoint (#153)
+- Clarify agent/sampler address overrides in README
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,9 +1,18 @@
...
-- Nothing yet
+- Improve C++ flag handling in CMake (#128)
+- localIP: try resolving local hostname first (#130)
+- Fix deadlock if steady_clock::now() returns the same value twice (#132)
+- Implement SpanContext::Clone (#138)
+- Change sampler param sentinel value from YAML parser (#145)
+- Fix RemoteReporter test race condition (#135)
+- Add details on how to set agent & collector endpoint (#153)
+- Clarify agent/sampler address overrides in README
+- Upgrade to OpenTracing 1.5.0; Support build on Windows (#115)
```
yurishkuro: this should certainly deserve a bump to 0.5.0, not a patch

mdouaihy: indeed, I hesitated bumping the version between choosing a name for the branch and editing the changelog.

mdouaihy: change pushed.

yurishkuro:
What about the other change `project(jaegertracing VERSION 0.5.0)`?

Usually we have all clients include their versions as span.process.tags, but it's more difficult with clients that are only released via tag (like Go), so in those we usually have a constant that needs to be bumped as part of the release.

mdouaihy:
This is done automatically at build time via CMake

```
static constexpr auto kJaegerClientVersion = "C++-@PROJECT_VERSION@";
```

yurishkuro:
0.5.0 released

PR #170, opened 2019-09-08
Back to development

yurishkuro:


CMakeLists.txt
```diff
@@ -19,7 +19,7 @@ HunterGate(
     SHA1 "29f10683f10c7b35e1f599d71542af0c2daa6a01"
 )
 
-project(jaegertracing VERSION 0.5.0)
+project(jaegertracing VERSION 0.5.1dev)
```
yurishkuro: @mdouaihy this didn't work for me, hunter complained
```
CMake Error at CMakeLists.txt:22 (project):
  VERSION "0.5.1dev" format invalid.
```
Is there another way to specify that the version is not "official"?

mdouaihy: @yurishkuro, this is CMake complaining about the version of the project and not Huinter.

According to the [documentation](https://cmake.org/cmake/help/v3.15/variable/CMAKE_VERSION.html), "0.5.1dev" is not a valid version.

mdouaihy:
@yurishkuro, I can find a CMake compatible pattern. Is there any pattern you have in mind?

CMakeLists.txt
yurishkuro: According to docs:

> Development versions of CMake use the format: `<major>.<minor>.<date>[-<id>]`
> where the <date> component is of format CCYYMMDD and <id> may contain arbitrary text. 

it looks like `0.5.20190908-dev` should be valid, but it still fails

yurishkuro:
In the other clients we typically use something like 0.5.1-dev, i.e. a version that makes it clear what is the _previous_ official release that this builds on top of, and that also indicates that this is a dev version, not an official release. Since the client versions are captured in the span tags, we often use that to get a sense of which client versions are being used in a given installation.

mdouaihy:
hello @yurishkuro, please check the commit https://github.com/mdouaihy/jaeger-client-cpp/commit/9b732a3a325a289da1ce1d26a245d90fc994fdd1.

The idea is to define a CMake variable BRANCH_STATUS and use it in the Constant.in file

mdouaihy:
the thing is that the project setting has a specific pattern and we need an additional variable for that.

yurishkuro:
looks reasonable

yurishkuro:
thanks

mdouaihy:
Hi @yurishkuro,

can we cancel this PR as it's irrelevant now?

Issue #121, opened 2018-08-06
Remove Thrift headers from Jaeger public headers


sebastian-garofalo:
Hi

The example app works, although one of the unit tests fails, but then when I try to use the library in my projects I get the following error.

```
In file included from /usr/local/include/jaegertracing/Tag.h:20:0,
                 from /usr/local/include/jaegertracing/LogRecord.h:20,
                 from /usr/local/include/jaegertracing/Span.h:26,
                 from /usr/local/include/jaegertracing/UDPTransport.h:20,
                 from /usr/local/include/jaegertracing/reporters/Config.h:25,
                 from /usr/local/include/jaegertracing/Config.h:23,
                 from /usr/local/include/jaegertracing/Tracer.h:28,
                 from followersServer.cc:11:
/usr/local/include/jaegertracing/thrift-gen/jaeger_types.h:12:27: fatal error: thrift/Thrift.h: No such file or directory
compilation terminated.
```
So I tried to update the thrift submodule as you mention in the README but doesn't seems to be working fine, I get:

```
sgarofalo@AR-IT08661:~/dev/jaeger-client-cpp$ find idl/thrift/ -type f -name \*.thrift -exec thrift -gen cpp -out src/jaegertracing/thrift-gen {} \;
[WARNING:/home/sgarofalo/dev/jaeger-client-cpp/idl/thrift/jaeger.thrift:18] No generator named 'netcore' could be found!
[WARNING:/home/sgarofalo/dev/jaeger-client-cpp/idl/thrift/zipkincore.thrift:19] No generator named 'netcore' could be found!
[WARNING:/home/sgarofalo/dev/jaeger-client-cpp/idl/thrift/agent.thrift:21] No generator named 'netcore' could be found!
[WARNING:/home/sgarofalo/dev/jaeger-client-cpp/idl/thrift/aggregation_validator.thrift:18] No generator named 'netcore' could be found!
[WARNING:/home/sgarofalo/dev/jaeger-client-cpp/idl/thrift/baggage.thrift:18] No generator named 'netcore' could be found!
[WARNING:/home/sgarofalo/dev/jaeger-client-cpp/idl/thrift/crossdock/tracetest.thrift:18] No generator named 'netcore' could be found!
[ERROR:/home/sgarofalo/dev/jaeger-client-cpp/idl/thrift/crossdock/tracetest.thrift:28] (last token was 'Downstream')
Type "Downstream" has not been defined.
[...]

PR #172, opened 2019-09-16
Remove Thrift headers from Jaeger public headers

BeDeBaMu:
- Adresses https://github.com/jaegertracing/jaeger-client-cpp/issues/121
so that thrift is a private dependency of jaeger.

- Moved the minimal set of code from .h to .cpp to avoid thrift-gen headers to be exposed as part of the API.

Signed-off-by: Benoit De Backer <bdebacker@murex.com>

### DIFF
--- a/src/jaegertracing/LogRecord.cpp
+++ b/src/jaegertracing/LogRecord.cpp
@@ -15,3 +15,25 @@
  */
 
 #include "jaegertracing/LogRecord.h"
+#include "jaegertracing/thrift-gen/jaeger_types.h"
+
+namespace jaegertracing {
+void LogRecord::thrift(thrift::Log& log) const
+{
+    log.__set_timestamp(std::chrono::duration_cast<std::chrono::microseconds>(
+                            _timestamp.time_since_epoch())
+                            .count());
+
+    std::vector<thrift::Tag> fields;
+    fields.reserve(_fields.size());
+    std::transform(std::begin(_fields),
+                   std::end(_fields),
+                   std::back_inserter(fields),
+                   [](const Tag& tag) {
+                       thrift::Tag thriftTag;
+                       tag.thrift(thriftTag);
+                       return thriftTag;
+                   });
+    log.__set_fields(fields);
+}
+}  // namespace jaegertracing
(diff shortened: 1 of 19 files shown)


yurishkuro:
@mdouaihy could you review this?

BeDeBaMu:
Hi,

It’s a *minimal* fix that should solve the primary concern.
However,  to reduce the number of exported headers (nearly 100) I’d rather add some new changes:
1. Adapt some code to avoid exposing implementation details
     a. Samplers
     b. Metrics
     c.	[optional] utils/YAML.h
2. change the code layout a little bit [To be discussed, there may be constraints I am not aware of]
     a. Move the *public* headers under one directory include/ (keep private ones under src)
3. Modify CMakeLists in order to
     a. export only public headers,
     b. use PRIVATE for lib when applicable 
     c. [bonus] improve project visualization in IDEs
4. Abstract thrift and remove it from Span, Tag, LogRecord, etc.

Benefits expected: 
-	from 1. & 3.a much less header exports 
-	from 2. easier CMakeLists code to export public headers, easier public include visibility
-	from 3.b easier linkage with jaeger-client-cpp
-	from 4 easier switch to another serializer (protobuf)

Do you have any particular concerns regarding these proposals? 
I will insert issues later so that we can discuss them.

mdouaihy:
Hi @yurishkuro, I just want you to know that @BeDeBaMu and I are from the same organisation. Before I validate, I want to check that it's not a problem for you.

BeDeBaMu:
> @yurishkuro I think that the change is ok but we need to continue the work to remove any reference to thrift completely from any public object (header, classed) as suggested.
@mdouaihy this is the purpose of my 4th item

yurishkuro:
Any way to see how this change impacts https://github.com/envoyproxy/envoy/tree/master/examples/jaeger-native-tracing ?

mdouaihy:
I just tested manually  and it's working.

However; I noticed that the build-plugin script needs to be updated to build a version linking statically it's dependencies. I ll add an issue for that.

yurishkuro:
does this resolve #121 ?

PR #180, opened 2019-10-02
Do not allocate 65k buffers for each span

dqminh:
The current code allocates a few 65k buffer (i.e. maxPacketSize) when writing out
thrift::Span which is wasteful, as single span rarely reaches this size. We
noticed a significant memory increase when running under AddressSanitizer.

Instead, use the default 1024 bytes buffer and relies on buffer's growth to deal
with large span.

Signed-off-by: Daniel Dao <dqminh@cloudflare.com>

### DIFF
--- a/src/jaegertracing/ThriftSender.cpp
+++ b/src/jaegertracing/ThriftSender.cpp
@@ -40,10 +40,10 @@ namespace {
 constexpr auto kEmitBatchOverhead = 30;
 
 template <typename ThriftType>
-int calcSizeOfSerializedThrift(apache::thrift::protocol::TProtocolFactory& factory, const ThriftType& base, int maxPacketSize)
+int calcSizeOfSerializedThrift(apache::thrift::protocol::TProtocolFactory& factory, const ThriftType& base)
 {
     std::shared_ptr<apache::thrift::transport::TMemoryBuffer> buffer(
-        new apache::thrift::transport::TMemoryBuffer(maxPacketSize));
+        new apache::thrift::transport::TMemoryBuffer());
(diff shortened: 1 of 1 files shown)


DavitSo:
Hi all is this will reduce memory usage in Collector?

mdouaihy:
@DavitSo, do you mean the Jaeger Collector?

DavitSo:
Yes

DavitSo:
Collector received X size Span and store in back end DB(if i'm wrong correct me), If I'm correct does this  affect on this DB size growing.
Thanks,

mdouaihy:
This is an optimisation in the client:
- For now, the thrift sender create a huge buffer at each serialisation of a Span.
- With the change, the buffer created will be of smaller change. This will reduce the memory allocated for the serialisation
- With my suggestion, I am proposing to allocated once the buffer and reuse it.

This should have zero impact on data sent to the Collector.

DavitSo:
ok , thanks for clarification!

yurishkuro:
Reusing the buffer is certainly the right approach, as long as concurrency concerns are taken into account. If it's possible that the sender could be invoked concurrently, then pooling of the buffers could be an option.

mdouaihy:
the sender is not concurrent so the resuing the buffer should be ok

dqminh:
@mdouaihy @yurishkuro apologized for long delay as i took a vacation :D 

I sent another change at https://github.com/jaegertracing/jaeger-client-cpp/pull/185 which reuse the buffer for a sender to calculate the individual thrift sizes, hopefully that's correct.

yurishkuro:
fixed in #185

PR #190, opened 2019-11-20
Add jaeger-debug-id as a tag

### LINKED ISSUE #136: Automatically add tag with value of jaeger-debug-id header
## Add  jaeger-debug-id header value as tag in C++ client

We're using Envoy and have configured it to use native Jaeger tracing.  If we force a trace by sending a jaeger-debug-id header it is very hard/impossible to find the trace in the Jaeger UI. The C++ client does not set a a tag with the value of the jaeger-debug-id header.  The use case is that as a developer I'd like to be able to initiate an operation from the browser, and then find the corresponding trace and see what services were involved in handling the request.  

The docs for the Go client say: 
> When Jaeger sees this header in the request that otherwise has no tracing context, it ensures that the new trace started for this request will be sampled in the "debug" mode (meaning it should survive all downsampling that might happen in the collection pipeline), and the root span will have a tag as if this statement was executed:
> span.SetTag("jaeger-debug-id", "some-correlation-id")

 AFAICT the debug header is partially supported by the C++ client.  When the debug header is present in the incoming headers, Jaeger C++ client sets the debug and sampled flags on the SpanContext so the trace does get written out, but it does not set a tag with the value of the debug header.  On an even slightly busy system it is pretty much impossible to find the trace in the Jaeger UI.

Set a tag with name "jaeger-debug-id" with value of the jaeger-debug-id header before reporting the span.

FWIW, I did attempt to fix this myself.  I can build Envoy and libjaegertracing_plugin.so and debug them, and guessed my way through the code, but I could never get the library I built to actually trace anything from Envoy.  My C++ is so rusty that I doubt if any PR from me would be accepted anyway.

In Tracer.cpp Tracer::StartSpanWithOptions I think it is necessary to pass the debug ID from the parent SpanContext here (eg: add the debugId parameter after StrMap()):
            ```ctx = SpanContext(traceID, spanID, parentID, flags, StrMap(), _debugId_);```

and set a tag, perhaps in Span::FinishWithOptions() conceptually like:
```
if ( !_context.debugID().empty() ) {
      // TODO: get the configured debug header?
      SetTag( "jaeger-debug-id", _context.debugID());
    } 
```

Note that the name of the debug header is configurable in the C++ client.  Perhaps the name of the tag should be the same as the configured header name.

goller:

resolves #136 

If the trace was started with a debugID, adding jaeger-debug-id
as tag ensures that the trace will be sampled in the "debug" mode.

Additionally, the root span will have this ID as a searchable tag.

### DIFF
--- a/src/jaegertracing/Tracer.cpp
+++ b/src/jaegertracing/Tracer.cpp
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+#include "jaegertracing/Tag.h"
 #include "jaegertracing/Tracer.h"
 #include "jaegertracing/Reference.h"
 #include "jaegertracing/TraceID.h"
@@ -85,6 +86,7 @@ Tracer::StartSpanWithOptions(string_view operationName,
                 flags |=
                     (static_cast<unsigned char>(SpanContext::Flag::kSampled) |
                      static_cast<unsigned char>(SpanContext::Flag::kDebug));
+                samplerTags.push_back(Tag(kJaegerDebugHeader, parent->debugID()));
             }
             else {
                 const auto samplingStatus =
--- a/src/jaegertracing/TracerTest.cpp
+++ b/src/jaegertracing/TracerTest.cpp
@@ -118,7 +118,7 @@ absTimeDiff(const typename ClockType::time_point& lhs,
 TEST(Tracer, testTracer)
 {
     {
-    
+
     const auto handle = testutils::TracerUtil::installGlobalTracer();
     const auto tracer =
         std::static_pointer_cast<Tracer>(opentracing::Tracer::Global());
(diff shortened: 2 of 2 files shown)


yurishkuro:
please use `resolves #123` text in the description so that GH can auto-close the ticket on merge

src/jaegertracing/Tracer.cpp
```diff
@@ -85,6 +86,7 @@ Tracer::StartSpanWithOptions(string_view operationName,
                 flags |=
                     (static_cast<unsigned char>(SpanContext::Flag::kSampled) |
                      static_cast<unsigned char>(SpanContext::Flag::kDebug));
+                samplerTags.push_back(Tag(kJaegerDebugHeader, parent->debugID()));
```
yurishkuro: is there a unit test which can check that?

goller: Hey @yurishkuro thanks for the review; tests added in 7d81527

goller:
Woot! Thanks !

PR #191, opened 2019-11-22
fix tiny typo

The-Alchemist:


yurishkuro:
please fix DCO

mdouaihy:
Hi @The-Alchemist, would you please fix the DCO so that we can merge your fix?

PR #200, opened 2020-01-24
Fix loading of sampling endpoint address from JAEGER_SAMPLING_ENDPOINT environment variable

ecourreges-orange:
fix conversion to double of JAEGER_SAMPLER_PARAM
rename env var JAEGER_SAMPLER_MANAGER_HOST_PORT to JAEGER_SAMPLER_SERVER_URL
properly load JAEGER_SAMPLER_SERVER_URL

Signed-off-by: CI-Bot for Emmanuel Courreges <emmanuel.courreges@orange.com>

### DIFF
--- a/src/jaegertracing/samplers/Config.cpp
+++ b/src/jaegertracing/samplers/Config.cpp
@@ -36,10 +36,14 @@ void Config::fromEnv()
     const auto param = utils::EnvVariable::getStringVariable(kJAEGER_SAMPLER_PARAM_ENV_PROP);
     if (!param.empty()) {
         std::istringstream iss(param);
-        int paramVal = 0;
+        double paramVal = 0;
         if (iss >> paramVal) {
             _param = paramVal;
         }
+    }
+    const auto samplingServerURL = utils::EnvVariable::getStringVariable(kJAEGER_SAMPLER_SERVER_URL_ENV_PROP);
+    if (!samplingServerURL.empty()) {
+        _samplingServerURL = samplingServerURL;
     }
 }
 
--- a/src/jaegertracing/samplers/Config.cpp
+++ b/src/jaegertracing/samplers/Config.cpp
@@ -41,7 +41,7 @@ void Config::fromEnv()
             _param = paramVal;
         }
     }
-    const auto samplingServerURL = utils::EnvVariable::getStringVariable(kJAEGER_SAMPLER_SERVER_URL_ENV_PROP);
+    const auto samplingServerURL = utils::EnvVariable::getStringVariable(kJAEGER_SAMPLING_ENDPOINT_ENV_PROP);
     if (!samplingServerURL.empty()) {
         _samplingServerURL = samplingServerURL;
     }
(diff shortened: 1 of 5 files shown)


yurishkuro:
> JAEGER_SAMPLER_MANAGER_HOST_PORT to JAEGER_SAMPLER_SERVER_URL

what is the motivation for this?

ecourreges-orange:
> 
> 
> > JAEGER_SAMPLER_MANAGER_HOST_PORT to JAEGER_SAMPLER_SERVER_URL
> 
> what is the motivation for this?

In the yaml configuration it is named samplingServerURL, and the env variables for samplers started with JAEGER_SAMPLER, so I decided to rename it like that for consistency (and not JAEGER_SAMPLER_SAMPLING_SERVER_URL which is also redundant). Also it is really a URL and not a host:port.
In any case that env var was not used by the code, only declared, so I don't think the renaming will break anyone.
However if you prefer the term "manager", then it's the yaml that would need to be changed instead, with more impact to users as yaml used to be the only way of loading conf.

yurishkuro:
We had a discussion (https://github.com/jaegertracing/jaeger/issues/1971#issuecomment-570075734) about the naming, and the preferred name is `JAEGER_SAMPLING_ENDPOINT`.

ecourreges-orange:
Done renaming the env var `JAEGER_SAMPLING_ENDPOINT` as requested.

src/jaegertracing/samplers/Config.h
```diff
@@ -47,7 +47,7 @@ class Config {
 
     static constexpr auto kJAEGER_SAMPLER_TYPE_ENV_PROP = "JAEGER_SAMPLER_TYPE";
     static constexpr auto kJAEGER_SAMPLER_PARAM_ENV_PROP = "JAEGER_SAMPLER_PARAM";
-    static constexpr auto kJAEGER_SAMPLER_MANAGER_HOST_PORT_ENV_PROP = "JAEGER_SAMPLER_MANAGER_HOST_PORT";
+    static constexpr auto kJAEGER_SAMPLING_ENDPOINT_ENV_PROP = "JAEGER_SAMPLING_ENDPOINT";
```
yurishkuro: We cannot simply replace the name, it will break existing deployments. The old property needs to be supported as deprecated.

ecourreges-orange: Yes we can simply replace on the cpp client, because the env props are recent and the loading of this one was not coded yet, it was unused in the code, so if anyone was using it in their software, they would have noticed that it had no effect. Hence my code and unit tests that now make it work.

src/jaegertracing/samplers/Config.cpp
```diff
@@ -36,11 +36,15 @@ void Config::fromEnv()
     const auto param = utils::EnvVariable::getStringVariable(kJAEGER_SAMPLER_PARAM_ENV_PROP);
     if (!param.empty()) {
         std::istringstream iss(param);
-        int paramVal = 0;
+        double paramVal = 0;
         if (iss >> paramVal) {
             _param = paramVal;
         }
```
yurishkuro: Probably should log a warning if value cannot be parsed. But it's not in scope for this PR.

PR #203, opened 2020-02-04
print errors when environment variable conversions to int or double fail

ecourreges-orange:
Signed-off-by: CI-Bot for Emmanuel Courreges <emmanuel.courreges@orange.com>

Probably should log a warning if value cannot be parsed. But it's not in scope for this PR.
_Originally posted by @yurishkuro in https://github.com/jaegertracing/jaeger-client-cpp/pull/200_

Log failure information on std::cerr for every conversion error of environment variable to int or to double.
The config is loaded from env when we don't have a logger yet, that's why there is no alternative to std:cerr which was already used in similar cases.

### DIFF
--- a/src/jaegertracing/Config.cpp
+++ b/src/jaegertracing/Config.cpp
@@ -53,6 +53,9 @@ void Config::fromEnv()
                 std::getline(tagStream, tagValue, '=');
                 if (std::getline(tagStream, tagValue, '=')) {
                     // error, should be logged somewhere
+                    std::cerr << "wrong format in tag " << tagKey
+                    << " of env var " << kJAEGER_TAGS_ENV_PROP
+                    << ", it can not contain an =\n";
                 }
                 else {
                     _tags.emplace_back(tagKey, tagValue);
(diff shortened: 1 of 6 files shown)


mdouaihy:
Hi @ecourreges-orange,

Thank you for your efforts.

Instead of writing to an std::cerr, why not passing a logger that can be used to pass the relevant information?

ecourreges-orange:
> 
> 
> Hi @ecourreges-orange,
> 
> Thank you for your efforts.
> 
> Instead of writing to an std::cerr, why not passing a logger that can be used to pass the relevant information?

I thought about that and then saw that the logger is initialized after the conf is read, which means that the logger is not available yet in the fromEnv method.
At least that's what I remember.

ecourreges-orange:
> 
> 
> Hi @ecourreges-orange,
> 
> Thank you for your efforts.

@mdouaihy I updated the code and it is now mergeable following your recent commits on ConfigTest.

I don't want to go overboard here and introduce interface changes to the Config classes, and I think the logger part can be dealt later in another PR.
Thanks.

PR #206, opened 2020-02-17
An extension of enum opentracing::SpanReferenceType, for a new Span.

Alek86:
The PR is the result of this discussion: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/opentracing/92uhSCO-Tu8/D83rq-j4GQAJ

Basically, I need a way to force a traceId from user's code. Didn't see such way, had to write some dirty code in my copy of jaeger-client-cpp code.

Yuri Shkuro suggested to add a new jaeger specific reference type. He called it SELF, I decided to give a more specific name because of limitations (see below).

It is a bit similar to GO implementation (https://github.com/jaegertracing/jaeger-client-go/issues/397) but I had some limitations:
1) I couldn't copy whole SpanContext, because jaegertracing::Tracer::StartSpanWithOptions implementation was not completely clear to me. So I decided to copy only traceID and spanID
2) There is an implicit rule in jaegertracing::Tracer::StartSpanWithOptions: if it's a root span, its spanID is the traceID.low(). I'm not sure if it's safe to overrule, so I didn't.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,12 +108,19 @@ JAEGER_SAMPLING_ENDPOINT | The url for the remote sampling conf when using sampl
 JAEGER_TAGS | A comma separated list of `name = value` tracer level tags, which get added to all reported spans. The value can also refer to an environment variable using the format `${envVarName:default}`, where the `:default` is optional, and identifies a value to be used if the environment variable cannot be found
 
 ### SelfRef
-Jaeger Tracer supports an additional reference type call 'SelfRef'. This allows a caller to provide traceId and spanId for the root span.
-Must be the lone reference.
-Usage example:
-  jaegertracing::SpanContext spanContextWithUserIDs { {1, 2}, 3, 0, 0, jaegertracing::SpanContext::StrMap() }; // TraceId and SpanID must be != 0
-  auto span = opentracing::Tracer::Global()->StartSpan("tracedFunction1", {jaegertracing::SelfRef(&spanContextWithUserIDs)});
+Jaeger Tracer supports an additional reference type call 'SelfRef'.
+It returns an opentracing::SpanReference which can be passed to Tracer::StartSpan
+to influence the SpanContext of the newly created span. Specifically, the new span inherits the traceID
+and spanID from the passed SELF reference. It can be used to pass externally generated IDs to the tracer,
+with the purpose of recording spans from data generated elsewhere (e.g. from logs), or by augmenting the
+data of the existing span (Jaeger backend will merge multiple instances of the spans with the same IDs).
+Must be the lone reference, can be used only for root spans.
 
+Usage example:
+```
+  jaegertracing::SpanContext customCtx { {1, 2}, 3, 0, 0, jaegertracing::SpanContext::StrMap() }; // TraceId and SpanID must be != 0
+  auto span = tracer->StartSpan("spanName", { jaegertracing::SelfRef(&customCtx) });
+```
 
 ## License
 
--- a/README.md
+++ b/README.md
@@ -107,6 +107,14 @@ JAEGER_SAMPLER_PARAM | The sampler parameter (double)
 JAEGER_SAMPLING_ENDPOINT | The url for the remote sampling conf when using sampler type remote. Default is http://127.0.0.1:5778/sampling
 JAEGER_TAGS | A comma separated list of `name = value` tracer level tags, which get added to all reported spans. The value can also refer to an environment variable using the format `${envVarName:default}`, where the `:default` is optional, and identifies a value to be used if the environment variable cannot be found
 
+### SelfRef
+Jaeger Tracer supports an additional reference type call 'SelfRef'. This allows a caller to provide traceId and spanId for the root span.
+Must be the lone reference.
+Usage example:
+  jaegertracing::SpanContext spanContextWithUserIDs { {1, 2}, 3, 0, 0, jaegertracing::SpanContext::StrMap() }; // TraceId and SpanID must be != 0
+  auto span = opentracing::Tracer::Global()->StartSpan("tracedFunction1", {jaegertracing::SelfRef(&spanContextWithUserIDs)});
+
+
 ## License
 
 [Apache 2.0 License](./LICENSE).
--- a/examples/App.cpp
+++ b/examples/App.cpp
@@ -24,6 +24,11 @@ void tracedSubroutine(const std::unique_ptr<opentracing::Span>& parentSpan)
 
 void tracedFunction()
 {
+    // jaegertracing::SpanContext spanContextWithUserIDs { {111, 222}, 333, 0, 0, {} }; // TraceId and SpanID must be != 0
+    // auto span = opentracing::Tracer::Global()->StartSpan(
+    // 	"tracedFunction1", {jaegertracing::JaegerSpecific_UseTheseIDs(&spanContextWithUserIDs)});
+    // // spanID of span is 222, not 333, because it's a root span
+
     auto span = opentracing::Tracer::Global()->StartSpan("tracedFunction");
     tracedSubroutine(span);
 }
--- a/examples/App.cpp
+++ b/examples/App.cpp
@@ -25,7 +25,7 @@ void tracedSubroutine(const std::unique_ptr<opentracing::Span>& parentSpan)
 void tracedFunction()
 {
     //{
-    //jaegertracing::SpanContext spanContextWithUserIDs { {111, 222}, 333, 0, 0, {} }; // TraceId and SpanID must be != 0
+    //jaegertracing::SpanContext spanContextWithUserIDs { {111, 222}, 333, 0, 0, jaegertracing::SpanContext::StrMap() }; // TraceId and SpanID must be != 0
     //auto span = opentracing::Tracer::Global()->StartSpan(
     //    "tracedFunction1", {jaegertracing::SelfRef(&spanContextWithUserIDs)});
     //}
--- a/examples/App.cpp
+++ b/examples/App.cpp
@@ -24,10 +24,11 @@ void tracedSubroutine(const std::unique_ptr<opentracing::Span>& parentSpan)
 
 void tracedFunction()
 {
-    // jaegertracing::SpanContext spanContextWithUserIDs { {111, 222}, 333, 0, 0, {} }; // TraceId and SpanID must be != 0
-    // auto span = opentracing::Tracer::Global()->StartSpan(
-    // 	"tracedFunction1", {jaegertracing::JaegerSpecific_UseTheseIDs(&spanContextWithUserIDs)});
-    // // spanID of span is 222, not 333, because it's a root span
+    //{
+    //jaegertracing::SpanContext spanContextWithUserIDs { {111, 222}, 333, 0, 0, {} }; // TraceId and SpanID must be != 0
+    //auto span = opentracing::Tracer::Global()->StartSpan(
+    //    "tracedFunction1", {jaegertracing::SelfRef(&spanContextWithUserIDs)});
+    //}
 
     auto span = opentracing::Tracer::Global()->StartSpan("tracedFunction");
     tracedSubroutine(span);
--- a/examples/App.cpp
+++ b/examples/App.cpp
@@ -24,12 +24,6 @@ void tracedSubroutine(const std::unique_ptr<opentracing::Span>& parentSpan)
 
 void tracedFunction()
 {
-    //{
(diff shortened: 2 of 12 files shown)


yurishkuro:
I would prefer to stick with the same name `Self` and not diverge between different Jaeger clients. There is an option question about the sampling in the Go implementation: https://github.com/jaegertracing/jaeger-client-go/pull/411#issuecomment-587207687. I tend to think that your implementation is actually better, although I can see that both scenarios have merit.

FWIW, it's not a requirement that `spanID == traceID.low()`, just the typical implementation aimed at saving CPU cycles by not generating yet another random ID.

src/jaegertracing/Tracer.cpp
```diff
@@ -101,7 +110,7 @@ Tracer::StartSpanWithOptions(string_view operationName,
         }
         else {
             const auto traceID = parent->traceID();
-            const auto spanID = randomID();
+            const auto spanID = self ? self->spanID() : randomID();
```
yurishkuro: this looks inconsistent. How do you rationalize only overriding span id, but not the trace id?

I think I would propose that SelfRef is only used as a lone reference for root span.

examples/App.cpp
```diff
@@ -24,6 +24,12 @@ void tracedSubroutine(const std::unique_ptr<opentracing::Span>& parentSpan)
 
 void tracedFunction()
 {
+    //{
+    //jaegertracing::SpanContext spanContextWithUserIDs { {111, 222}, 333, 0, 0, jaegertracing::SpanContext::StrMap() }; // TraceId and SpanID must be != 0
+    //auto span = opentracing::Tracer::Global()->StartSpan(
+    //    "tracedFunction1", {jaegertracing::SelfRef(&spanContextWithUserIDs)});
+    //}
```
yurishkuro: remove?

src/jaegertracing/Reference.h
```diff
@@ -30,6 +30,9 @@ namespace thrift {
 class SpanRef;
 }
 
+// An extension of enum opentracing::SpanReferenceType, for a new Span. Only to copy traceID and spanID
+const static int SpanReferenceType_JaegerSpecific_SelfRef = 99;
```
yurishkuro: does this need to be in `.h`? It sounds like an implementation detail.

src/jaegertracing/Tracer.h
```diff
@@ -301,6 +303,14 @@ class Tracer : public opentracing::Tracer,
     int _options;
 };
 
+
+// jaegertracing::SelfRef returns a StartSpanOption pointing to the Span which traceID and spanID should become the new Span's IDs
+//
+// See opentracing::SpanReference
+inline opentracing::SpanReference SelfRef(const opentracing::SpanContext* span_context) noexcept {
```
yurishkuro: considering that there is already a custom type `Reference` (unlike in Go), I think it might be better to go with a static method on that class, `Reference::Self(spanCtx)`

src/jaegertracing/Tracer.cpp
Alek86: > I think I would propose that SelfRef is only used as a lone reference for root span.
What if the library user wants to generate all spanIDs by themselves? I think it would be inconsistent, to let users generate spanIDs for root spans, but not for children.

examples/App.cpp
Alek86: I wanted to leave this commented code as an example of SelfRef using. Will remove in a next commit.

src/jaegertracing/Reference.h
Alek86: Agree, will move to a .cpp or at least inside the Reference class (it has no .cpp)

src/jaegertracing/Tracer.h
Alek86: opentracing::SpanReference and opentracing::StartSpanOption are declared in opentracing/tracer.h which is not included in jaegertracing/Reference.h
I could do what you requested, but with adding a new - from my perspective unrelated - include. Please let me know if you still want it

yurishkuro: ok, let's keep it here. OpenTracing includes appear to be poorly thought through, e.g. `SpanReferenceType` is defined in `propagation.h`.

src/jaegertracing/Reference.h
Alek86: Done

examples/App.cpp
Alek86: Removed

src/jaegertracing/Tracer.cpp
Alek86: SelfRef can only be used for root spans, using them with any other reference will make StartSpanWithOptions return nullptr

README.md
```diff
@@ -107,6 +107,14 @@ JAEGER_SAMPLER_PARAM | The sampler parameter (double)
 JAEGER_SAMPLING_ENDPOINT | The url for the remote sampling conf when using sampler type remote. Default is http://127.0.0.1:5778/sampling
 JAEGER_TAGS | A comma separated list of `name = value` tracer level tags, which get added to all reported spans. The value can also refer to an environment variable using the format `${envVarName:default}`, where the `:default` is optional, and identifies a value to be used if the environment variable cannot be found
 
+### SelfRef
+Jaeger Tracer supports an additional reference type call 'SelfRef'. This allows a caller to provide traceId and spanId for the root span.
+Must be the lone reference.
+Usage example:
+  jaegertracing::SpanContext spanContextWithUserIDs { {1, 2}, 3, 0, 0, jaegertracing::SpanContext::StrMap() }; // TraceId and SpanID must be != 0
```
yurishkuro: use 3 back ticks for code formatting

yurishkuro: This is how this currently renders:
image

src/jaegertracing/Tracer.cpp
```diff
@@ -33,6 +33,10 @@ using SystemClock = Tracer::SystemClock;
 using SteadyClock = Tracer::SteadyClock;
 using TimePoints = std::tuple<SystemClock::time_point, SteadyClock::time_point>;
 
+// An extension of enum opentracing::SpanReferenceType, for a new Span. Only to copy traceID and (for non-root spans) spanID
+// spanID for root spans still traceID.low()
```
yurishkuro: Comment doesn't read as full sentences. It's not necessary here, can just say
```
// An extension of opentracing::SpanReferenceType enum. See SelfRef(). 
```

README.md
yurishkuro:
```suggestion
  jaegertracing::SpanContext customCtx { {1, 2}, 3, 0, 0, jaegertracing::SpanContext::StrMap() }; // TraceId and SpanID must be != 0
```

README.md
```diff
@@ -107,6 +107,14 @@ JAEGER_SAMPLER_PARAM | The sampler parameter (double)
...
+  auto span = opentracing::Tracer::Global()->StartSpan("tracedFunction1", {jaegertracing::SelfRef(&spanContextWithUserIDs)});
```
yurishkuro:
```suggestion
  auto span = tracer->StartSpan("spanName", { jaegertracing::SelfRef(&customCtx) });
```

src/jaegertracing/Tracer.h
```diff
@@ -301,6 +303,12 @@ class Tracer : public opentracing::Tracer,
     int _options;
 };
 
+
+// jaegertracing::SelfRef returns a StartSpanOption pointing to the Span which traceID and spanID should become the new Span's IDs
```
yurishkuro:
```
// jaegertracing::SelfRef returns an opentracing::SpanReference which can be passed to Tracer::StartSpan
// to influence the SpanContext of the newly created span. Specifically, the new span inherits the traceID
// and spanID from the passed SELF reference. It can be used to pass externally generated IDs to the tracer,
// with the purpose of recording spans from data generated elsewhere (e.g. from logs), or by augmenting the
// data of the existing span (Jaeger backend will merge multiple instances of the spans with the same IDs).

Comment (yurishkuro) [reply src/jaegertracing/Tracer.h]: I suggest copying the description to README as well.

Comment (Alek86) [reply src/jaegertracing/Tracer.h]: Updated with minor changes, added to README as well

Also added:
Must be the lone reference, can be used only for root spans

Please let me know if you don't need this line

File: src/jaegertracing/Tracer.cpp [hunk 9]
```diff
@@ -67,19 +70,32 @@ Tracer::StartSpanWithOptions(string_view operationName,
     try {
         const auto result = analyzeReferences(options.references);
         const auto* parent = result._parent;
+        const auto* self = result._self;
         const auto& references = result._references;
+        if (self && (parent || !references.empty()))
```
mdouaihy: When looking at the Go implementation, having a parent/references does not cause an error. They are simply ignored.

Is there any reason to do it differently here?

Alek86: 1)
About unit tests. I tried to add a UT a few times and it always ended with CI failing with an Access Denied during calling of StartSpan with ChildOf.
I'm 99.9% sure the reason is in dependencies (in opentracing.lib/.dll to be more precise) because I had very much the same problem when I tried to run the App.cpp on my Win10+VS 2019 computer.
I fixed it locally by building opentracing-cpp 1.5.x project and using its results in jaegertracing code.

To prove that my changes have nothing to do with the Access Denied, I created another temp PR where the only change is in TracerTest.cpp and received the same error: https://github.com/jaegertracing/jaeger-client-cpp/pull/208
I'm not familiar with Hunter, so cannot fix this problem myself.

That's the reason I couldn't add any changes in unittests.

----

2) 
About returning nullptr if SelfRef is used: I thought that if a library user is making a mistake of setting both parent span and selfref, the library should react with some type of mistake (in this case it would return a nullptr). IMO creating a root span instead of a requested child span in this case is closer to "unspecified behavior" than a reaction to an error.
If you think the latter behavior is better for the project, I'll reimplement it.

mdouaihy: I will check your test tomorrow as I ll have access to a Windows machine and will try to help you on that.

mdouaihy: Hi @Alek86, I tried your change locally on my Windows machine and indeed, stumbled on the same error.

I then updated the OpenTracing version to 1.6.0 and the exception was fixed. In  fact, there is a problem with the debug flags of OpenTracing that was fixed with the latest release. That bug was leading to different object layout between OpenTracing lib and Jaeger one, leading to that behavior.

I will start by updating OpenTracing dependency to 1.6.0.

Alek86: Thank you! I suppose I should wait for your fix (in master?), then merge it to this branch and then add the UT.
Please also let me know if you'll accept returning nullptr behavior or still want to ignore ChildOf etc if SelfRef is passed

mdouaihy: hello @Alek86, I opened the PR #209 to upgrade to OpenTracing 1.6.0. This should solve your crash under WIndows.

mdouaihy: Hello @Alek86, PR #209  was merged, You can rebase now and check the unit test.

Alek86: Hi @mdouaihy, I've added the unit tests

mdouaihy: Hi @Alek86, thank you for your effort.

One more think I just noticed, IMO, I think that if we throw an exception when the span state is invalid, maybe we can throw a typed exception with the relevant information.   std::exception() is too generic. What do you think?

Alek86: Okay. Do you have a place in mind where I should add the new exception class?

Alek86: Replaced std::exception with std::invalid_argument with explanation text
Added the explanation text to the error log

mdouaihy: I believe that the message should state that the Self reference and other types of references are exclusive. The current message informs only about self. What do you think?

Alek86: I think that "self reference must be the only reference" text also means that the other types cannot be used with self ref.
If you prefer to see another text, can you give it to me, I'll use it?

mdouaihy: I suggest: Self and references are exclusive. Only one can be specified. What do you think?

yurishkuro: Other references do not need to come as "only one".

Suggestion: "Self reference cannot be combined with other reference types"

Alek86: I'm ready for any text change that will help this code reach master :)
I'll add "Self and references are exclusive. Only one of them can be specified"

Alek86: Done

PR #208, opened 2020-02-22
Test UT

Alek86:
## Please ignore, I'll remove this PR soon.

Somehow any of my unittest changes from the other PR are failing the unittest CI step with a strange access violation, but the same unittests are working on my Win+VS machine after I rebuilt opentracing library.

So I added a same unittest change in a "clean" master code to see if it fails the same way. If it does, this will mean that there are some library dependency problem, not originated by me adding SelfRef.

Signed-off-by: Oleksandr Bukaiev <bukaevav@gmail.com>


Alek86:
This PR was to show a (probably .hunter related) problem, closing

Issue #141, opened 2018-12-13
Does jaeger-client-cpp support constructing the tracer data in X-B3 format and reporting it?


hafent:
Does jaeger-client-cpp support constructing the tracer data in X-B3 format and reporting it? Or can I use zipkin-cpp-opentracing to match jaeger to construct the tracer data in X-B3 format?

isaachier:
As of now it does not. Potentially will be supported in the future.

hafent:
Can I use this software combination ("opentracing-cpp"&& "zipkin-cpp-opentracing" && "nginx-opentracing" && "jaeger") to build and propagate X-B3 tracer data?

yurishkuro:
Many Jaeger clients provide B3 codecs that can be enabled on the tracer.  It may not have been implemented in C++ client.

Issue #125, opened 2018-08-17
Propagating B3-Headers


sebastian-garofalo:
Hi guys

I was able to create new spans on my POC C++ app, now  I want the spans to include the b3 headers so jaeger can correlate the span with its originating request trace.

I was doing something like this:

```
TracingData td = ExtractTraceData(context);
auto span = opentracing::Tracer::Global()->StartSpan("Test B3 in baggage");
span->SetBaggageItem(HEADER_X_REQUEST_ID, td.xRequestId);
span->SetBaggageItem(HEADER_X_TRACE_ID, td.xB3TraceId);
span->SetBaggageItem(HEADER_X_SPAN_ID, td.xB3SpanId);
```

Should I use the baggage for that? I'm seeing jaeger UI shows the trace ID as a "tag" but the stuff I put into the baggage doesn't appears to be a tag.

At opentracing gitter I was told I need to use a B3 codec and that maybe there isn't one for C++

Do you guys know how could I propagate the B3 headers in a span?

Thanks for your help

isaachier:
Trace ID is not represented as a tag. The UI might cause it to look like it is a tag, but it is not. I don't understand your example. What do you need the identifier for specifically? We handle the trace metadata internally and prefer not to expose it directly. Eventually, we will expose a method for this in the OpenTracing API, but not yet. Also, you are correct that we have no plans of supporting B3 at this time.

yurishkuro:
I think if someone creates a PR with B3 codec, we'll accept it, most other Jaeger clients support it.

isaachier:
Really? I gotta look into this some more...

yurishkuro:
Here's an example from Java client: https://github.com/jaegertracing/jaeger-client-java/blob/dfb9d8b9dc849f2e6f880b44b26fb77dfc894a02/jaeger-core/src/main/java/io/jaegertracing/internal/propagation/B3TextMapCodec.java

Pratik-Shah:
Hi All, we have a similar requirement where our applications are c++ based running along with service mesh ( istio ) . We started using jaeger client for the application without service mesh and post service mesh it seems we will be unable to extract the headers and create a childOf relation from the envoy span as there are based on zipkin.  My assumption here is that this feature of having B3 codec would help. So is there a plan to have it in c++ client. Or is there a different approach we should take ? Please guide.

isaachier:
[...]

PR #210, opened 2020-02-25
Adding x-b3-* headers parsing and propagating support

kunalekawde:
- #125 and probably  #141

- The changes are parse / decode the tracing information received in x-b3-traceId, x-b3-spanId, x-b3-parentspanId, x-b3-sampled and to propagate them further.

### DIFF
--- a/src/jaegertracing/SpanContext.h
+++ b/src/jaegertracing/SpanContext.h
@@ -32,6 +32,13 @@
 
 namespace jaegertracing {
 
+enum class TracerType
+{
+    TRACER_TYPE_UNKNOWN,
+    TRACER_TYPE_UBER,
+    TRACER_TYPE_B3,
+};
+
 class SpanContext : public opentracing::SpanContext {
   public:
     using StrMap = std::unordered_map<std::string, std::string>;
@@ -45,7 +52,7 @@ class SpanContext : public opentracing::SpanContext {
         , _spanID(0)
         , _parentID(0)
         , _flags(0)
-        , _b3Headers(false)
+        , _tracerType(TracerType::TRACER_TYPE_UNKNOWN)
         , _mutex()
     {
     }
@@ -62,7 +69,7 @@ class SpanContext : public opentracing::SpanContext {
         , _flags(flags)
         , _baggage(baggage)
         , _debugID(debugID)
-        , _b3Headers(false)
+        , _tracerType(TracerType::TRACER_TYPE_UNKNOWN)
         , _mutex()
     {
     }
@@ -73,14 +80,14 @@ class SpanContext : public opentracing::SpanContext {
                 unsigned char flags,
                 const StrMap& baggage,
                 const std::string& debugID,
-                bool b3Present)
+                TracerType type)
         : _traceID(traceID)
         , _spanID(spanID)
         , _parentID(parentID)
         , _flags(flags)
         , _baggage(baggage)
         , _debugID(debugID)
-        , _b3Headers(b3Present)
+        , _tracerType(type)
         , _mutex()
     {
     }
@@ -92,7 +99,7 @@ class SpanContext : public opentracing::SpanContext {
         , _flags(ctx._flags)
         , _baggage(ctx._baggage)
         , _debugID(ctx._debugID)
-        , _b3Headers(ctx._b3Headers)
+        , _tracerType(ctx._tracerType)
     {
     }
 
@@ -111,7 +118,7 @@ class SpanContext : public opentracing::SpanContext {
         swap(_flags, ctx._flags);
         swap(_baggage, ctx._baggage);
         swap(_debugID, ctx._debugID);
-        swap(_b3Headers, ctx._b3Headers);
+        swap(_tracerType, ctx._tracerType);
     }
 
     friend void swap(SpanContext& lhs, SpanContext& rhs) { lhs.swap(rhs); }
@@ -128,7 +135,7 @@ class SpanContext : public opentracing::SpanContext {
     {
         std::lock_guard<std::mutex> lock(_mutex);
         return SpanContext(
-            _traceID, _spanID, _parentID, _flags, baggage, _debugID, _b3Headers);
+            _traceID, _spanID, _parentID, _flags, baggage, _debugID, _tracerType);
     }
 
     template <typename Function>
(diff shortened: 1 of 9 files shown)


src/jaegertracing/SpanContext.h
```diff
@@ -45,6 +45,7 @@ class SpanContext : public opentracing::SpanContext {
         , _spanID(0)
         , _parentID(0)
         , _flags(0)
+        , _b3Headers(false)
```
yurishkuro: This doesn't belong here. Parsing of B3 headers is the job for codecs (see https://github.com/jaegertracing/jaeger-client-cpp/tree/master/src/jaegertracing/propagation). You can use Jaeger clients in other languages as a reference.

kunalekawde:
> 
> 
> You can use Jaeger clients in other languages as a reference.

Sure thanks, shall try that

kunalekawde:
@yurishkuro Needed some help
1. Can we assume that uber-ctx-* & x-b3-* would be mutually exclusive to any node ?
2. As I see in java implementation (https://github.com/jaegertracing/jaeger-client-java/tree/master/jaeger-core/src/main/java/io/jaegertracing/internal/propagation) the codecs are interpreted as text, http, binary and b3. If we have to same at jaeger cpp client wont it be a substantial change w.r.t class design ? Moreover B3 is an alternative to Uber-Ctx, does it fit to be in classification of codes inline to text/http/binary ? Wouldn't B3 also ever need binary/text ?. 
Please let me know your comments which would help me further.

yurishkuro:
1 - yes, I would treat them as mutually exclusive. If someone really wants to support both simultaneously, they can implement another delegating codec on top of those two. 

2- text, http, and binary are the three formats defined in opentracing. B3 would only implement text and http, probably with the same codec since b3 syntax is http-safe (unlike Jaeger’s baggage that requires url-encoding for values in http).

kunalekawde:
> 2- text, http, and binary are the three formats defined in opentracing. B3 would only implement text and http, probably with the same codec since b3 syntax is http-safe (unlike Jaeger’s baggage that requires url-encoding for values in http).
 So can we have B3 at level of Uber-ctx type - probably abstracting common layer and then provide text/http/other encoding to it. Basically unlike java implementation then B3 wont be a codec which implements extract and inject but a user of it.

yurishkuro:
If there's a common layer to abstract, I'm for it, but in the other clients I didn't see that, as the core of the codecs is dealing with specific headers of a given context format.

src/jaegertracing/SpanContext.h
```diff
@@ -45,6 +52,7 @@ class SpanContext : public opentracing::SpanContext {
         , _spanID(0)
         , _parentID(0)
         , _flags(0)
+        , _tracerType(TracerType::TRACER_TYPE_UNKNOWN)
```
yurishkuro: Why is this needed? The usual approach is to configure tracer with one type of codecs, which apply to both inbound and outbound requests. There’s no need to remember what type of encoding was used in the inbound.

kunalekawde: But how and where do we need to configure a particular type of codec and based on what?
Here in this PR, I've based it on the incoming request headers. As it would be either B3 / Uber, one of the tracer type would be set. If the trace is initiated by that node itself, it would be set with options in startSpanWithOptions API. 
If we dont remember type of tracing then using injectSpan, we need to have some flag to indicate what type of tracing should be done so that appropriate headers can be set.

yurishkuro: See https://github.com/jaegertracing/jaeger-client-java/pull/694/files, specifically `CodecConfiguration.withPropagation`

kunalekawde: @yurishkuro thanks for the reference, I still fail to understand on how this implementation can be used dynamically. As per readme:
```
b3Codec = new B3TextMapCodec();
tracer = new JaegerTracer.Builder(serviceName)
  .registerInjector(Format.Builtin.HTTP_HEADERS, b3Codec)
  .registerExtractor(Format.Builtin.HTTP_HEADERS, b3Codec)
  ...
```
I also checked in stagemonitor.git, looks like application is/have to build a object at start by specifying the tracing type e.g. B3 / Uber. 

If this is so then, application is aware what kind of headers it is expecting and there wouldn't be a need to remember the tracing type. 

If I'm missing something, please help point it out.

yurishkuro: The code snippet you showed is exactly how you do it. The tracer is created once at application start-up, and at that time you can replace the codecs. After that no more "dynamic" configuration is needed, nor does the span need to remember which codec is used.

kunalekawde: oh ok, which means that application has to be aware of what type tracing type has to supported explicitly which can be either Uber or B3 mutually exclusively.

yurishkuro: Yes, but the type of context used on the wire is usually the same for the whole site, not varied per application or per request.

It is theoretically possible to write a composite codec combined from both Uber and B3 headers, if it's required for backwards compatibility. But a single format for the whole system is highly recommended.

kunalekawde: Sure thanks, I had made the changes to handle kind of dynamic selection expecting either Uber or B3 headers at runtime at the node. An API getTracerType checks for incoming headers and uses particular tracer type - Uber/B3.

yurishkuro: Sorry, but I don't think we want to support this "dynamic" behavior. None of the other Jaeger clients do it this way, and I have not seen anyone ever asking for this. The codec is configured when creating a tracer, it does not change after that. A composite codec can be implemented ([example in java](https://github.com/jaegertracing/jaeger-client-java/blob/dfb9d8b9dc849f2e6f880b44b26fb77dfc894a02/jaeger-core/src/main/java/io/jaegertracing/internal/propagation/CompositeCodec.java)), which handles more than one encoding, but it is also doing them _always_, not dynamically. As a result, there's never a need to store "type" in the span context.

I don't know why you're trying to reinvent the wheel. I pointed you to several examples how codecs are done in Jaeger clients in other languages, it's an established pattern.

kunalekawde: Not being certain of the ingress nodes and its supported types(e.g. Istio currently supports only B3) we thought being flexible to handle any type based on incoming headers would be good. 
Yes, as per the case you mentioned of static provisioning this PR would need more enhancement. Thanks for all the help and references. I shall probably create a new one and close this.

kunalekawde:
closing this PR to have static provisioning of tracing types for which PR needs to be enhanced.

PR #219, opened 2020-03-20
add Findnlohmann_json

ideepika:
even after json package installed using distro package, cmake
is not able to search for it's cmake config, adding a find_package
method ensures garanteed check on finding dependency, which is either
skipped or not discovered.

Reference:

CMake Error at CMakeLists.txt:80 (find_package):
  Could not find a package configuration file provided by "nlohmann_json"
  with any of the following names:

    nlohmann_jsonConfig.cmake
    nlohmann_json-config.cmake

  Add the installation prefix of "nlohmann_json" to CMAKE_PREFIX_PATH or set
  "nlohmann_json_DIR" to a directory containing one of the above files.  If
  "nlohmann_json" provides a separate development package or SDK, be sure it
  has been installed.

Adapted from: https://github.com/jaegertracing/jaeger-client-cpp/pull/46
Signed-off-by: Deepika Upadhyay <dupadhya@redhat.com>

-

-

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -83,7 +83,7 @@ endif()
 list(APPEND package_deps OpenTracing)
 
 hunter_add_package(nlohmann_json)
-find_package(nlohmann_json CONFIG REQUIRED)
+find_package(nlohmann_json 2.1.0 ${hunter_config} REQUIRED)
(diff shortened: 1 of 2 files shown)


ideepika:
@mdouaihy

PR #222, opened 2020-05-03
remove gcc version check

fvahid:
-

-

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,12 +53,6 @@ else()
 endif()
 
 set(package_deps)
-
-if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND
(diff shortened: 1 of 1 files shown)


joconnor-openet:
I think this risks breaking a nunch of functionality because of this bug with std::regex in gcc 4.8.x: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53631

I was thinking of submitting a pull request to fall back to boost::regex instead for gcc 4.8.x...

fvahid:
@joconnor-openet , thanks for you reply but as you know we don't have good solution to install gcc-4.9 on centos 7 . and I have alot of trouble to build jaeger-client-cpp into there, I thinks we this check we lost centos user to use this repo

Issue #179, opened 2019-10-01
Is gcc >= 4.9 requirement is really mandatory?


defanator:
## Requirement - what kind of business use case are you trying to solve?

Build a working combo of nginx + opentracing-nginx + jaeger tracer plugin on CentOS 7.6.

## Problem - what in Jaeger blocks you from solving the requirement?

Jaeger tracer has the requirement of having gcc >= 4.9:
https://github.com/jaegertracing/jaeger-client-cpp/blob/master/CMakeLists.txt#L56-L59

CentOS 7.6 has 4.8.5.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

I've removed the check for gcc >= 4.9, and jaeger tracer builds just fine. Here's the platform details:

```
# cat /etc/os-release 
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
PRETTY_NAME="CentOS Linux 7 (Core)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:7"
HOME_URL="https://www.centos.org/"
BUG_REPORT_URL="https://bugs.centos.org/"

CENTOS_MANTISBT_PROJECT="CentOS-7"
CENTOS_MANTISBT_PROJECT_VERSION="7"
REDHAT_SUPPORT_PRODUCT="centos"
REDHAT_SUPPORT_PRODUCT_VERSION="7"

# gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/libexec/gcc/x86_64-redhat-linux/4.8.5/lto-wrapper
Target: x86_64-redhat-linux
[...]

PR #223, opened 2020-05-07
Fall back to boost::regex when compiling under gcc <4.9

joconnor-openet:
, because std::regex is broken under gcc-4.8.x (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53631)

Signed-off-by: Joe O'Connor <Joe.OConnor@openet.com>

- Resolves: #179 

- CMakeLists.txt: define USE_BOOST_REGEX and add boost_regex.so dependency if gcc < 4.9, instead of failing the build
- add boost::regex code wrapped inside "#ifdef USE_BOOST_REGEX" 
-- this means the existing std::regex code should be unaffected with gcc>=4.9

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,9 +54,15 @@ endif()
 
 set(package_deps)
 
+# fall back to boost::regex for gcc 4.8.x and earlier
+# (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53631)
+set(BOOST_MIN_VERSION 1.55)
 if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND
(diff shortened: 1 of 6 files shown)


yurishkuro:
According to https://gcc.gnu.org/releases.html, gcc 4.9 was released in 2014. My personal opinion (non-blocking) is that it's not worth having so much spaghetti code in so many places just to support 6yrs old version of the compiler. At minimum, it should be better factored out into a single util file so that the rest of the code is not exposed to these `#if-#else`. Also, if we decide to explicitly add support for earlier gcc versions, then it needs to be added to CI as an option.

joconnor-openet:
Created new pull req https://github.com/jaegertracing/jaeger-client-cpp/pull/224 incorporating review comments, closing this one

PR #224, opened 2020-05-08
Fall back to boost::regex when building with gcc <4.9

### LINKED ISSUE #179: Is gcc >= 4.9 requirement is really mandatory?

joconnor-openet:
Needed because std::regex is broken in gcc-4.8.x
-- https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53631
This change fixes the library on RedHat 7.x, which uses gcc-4.8.5.

Signed-off-by: Joe O'Connor <Joe.OConnor@openet.com>

- Resolves: #179

- removed the #ifdefs from the code as suggested in previous review comment
- instead there is now a utils/Regex.h wrapper

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,17 @@ matrix:
         - ubuntu-toolchain-r-test
         packages:
         - *1
+        - g++-4.8
+    env:
+    - MATRIX_EVAL="CC=gcc-4.8 && CXX=g++-4.8 && CMAKE_OPTIONS='-DCMAKE_BUILD_TYPE=Debug
+      -DBUILD_SHARED_LIBS=ON -DJAEGERTRACING_COVERAGE=ON'"
+  - os: linux
+    addons:
+      apt:
+        sources:
+        - ubuntu-toolchain-r-test
+        packages:
+        - *1
         - g++-4.9
     env:
     - MATRIX_EVAL="CC=gcc-4.9 && CXX=g++-4.9 && CMAKE_OPTIONS='-DCMAKE_BUILD_TYPE=Debug
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,7 +56,10 @@ set(package_deps)
 
 if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND
    CMAKE_CXX_COMPILER_VERSION VERSION_LESS "4.9")
-  message(FATAL_ERROR "Must use gcc >= 4.9")
+  add_definitions(-DUSE_BOOST_REGEX)
+  hunter_add_package(Boost COMPONENTS regex)
+  find_package(Boost CONFIG REQUIRED regex)
+  list(APPEND LIBS "${Boost_REGEX_LIBRARY_RELEASE}")
 endif()
 
 hunter_add_package(thrift)
--- a/src/jaegertracing/net/URI.cpp
+++ b/src/jaegertracing/net/URI.cpp
@@ -15,12 +15,12 @@
  */
 
 #include "jaegertracing/net/URI.h"
+#include "jaegertracing/utils/Regex.h"
 
 #include <cassert>
 #include <cstring>
 #include <iomanip>
 #include <iostream>
-#include <regex>
 #include <cctype>
 
 namespace jaegertracing {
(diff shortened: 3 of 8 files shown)


yurishkuro (commented):
please add a respective gcc version to CI matrix .travis.yml

yurishkuro (approved):
lgtm.

@mdouaihy any concerns?

.travis.yml
```diff
@@ -7,6 +7,17 @@ addons:
...
+    addons:
+      apt:
+        sources:
+        - ubuntu-toolchain-r-test
+        packages:
+        - *1
+        - g++-4.8
+    env:
+    - MATRIX_EVAL="CC=gcc-4.8 && CXX=g++-4.8 && CMAKE_OPTIONS='-DCMAKE_BUILD_TYPE=Debug
+      -DBUILD_SHARED_LIBS=ON -DJAEGERTRACING_COVERAGE=ON'"
```
yurishkuro, line 20: seems to be some indentation issue

mdouaihy:
hi @joconnor-openet,
the PR seems ok for me. Thank you for you efforts.

however, @yurishkuro, I have a small concern about the old compilers support: This adds constraints and prevents from using newer C++ features. the regex here is an example. As you can see, the change is not minor: it's few lines of code but a whole library was needed. While this is ok here, this solution can't be generalised for everything else.
   
On one hand, I know that this is a constraint of libraries like ours. On the other hand, this prevents enhancements. Maybe a compiler depreciation strategy will be needed.

yurishkuro:
@mdouaihy I feel similar about supporting legacy compiler versions. 4.9 was released in 2014, there was plenty of time for people to upgrade. How do other widespread C++ libraries deal with this issue? Can we borrow the compiler support strategy from one of them?

Side note: I am frankly surprised by the need to use regex to deal with URLs! Most languages I dealt with since starting Jaeger had URL support built-in into std lib.

joconnor-openet:
Hi, is it possible to get this pull request merged, or do you want to adopt a different approach?
Thanks
Joe

PR #254, opened 2020-11-26
Support configuration for generating 128 bit trace ids

tobiasstadler:
- Using 128 Bit trace ids increases interoperability with W3C Trace Context (and other Jaeger clients)
- Resolves #254

- added a flag if 128 bit trace ids should be generated

### DIFF
--- a/src/jaegertracing/Config.cpp
+++ b/src/jaegertracing/Config.cpp
@@ -32,6 +32,12 @@ void Config::fromEnv()
         _disabled = disabled.second;
     }
 
+    const auto traceId128Bit =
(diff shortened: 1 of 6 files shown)


yurishkuro:
To clarify, the PR is adding _configuration_ support for 128bit IDs, correct? Not the actual support in the tracer.

tobiasstadler:
Yes, I only added configuration support. Sorry, the title is a bit misleading.

tobiasstadler:
I updated the title. Should I also change the commit message?

yurishkuro:
Thanks. The PR title is enough, we squash the commits during merge.

yurishkuro (approved):

PR #259, opened 2021-01-29
Do not strip leading zeros from trace IDs and span IDs

tobiasstadler:
- Implements jaegertracing/jaeger#1657

### DIFF
--- a/src/jaegertracing/TraceID.h
+++ b/src/jaegertracing/TraceID.h
@@ -43,8 +43,13 @@ class TraceID {
     template <typename Stream>
     void print(Stream& out) const
     {
-        out << std::setw(16) << std::setfill('0') << std::hex << _high
-            << std::setw(16) << std::setfill('0') << std::hex << _low;
+        if (_high == 0) {
+            out << std::setw(16) << std::setfill('0') << std::hex << _low;
+        }
+        else {
+            out << std::setw(16) << std::setfill('0') << std::hex << _high
+                << std::setw(16) << std::setfill('0') << std::hex << _low;
+        }
     }
 
     uint64_t high() const { return _high; }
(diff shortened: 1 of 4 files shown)


src/jaegertracing/TraceID.h
```diff
@@ -43,13 +43,8 @@ class TraceID {
...
     void print(Stream& out) const
     {
-        if (_high == 0) {
-            out << std::hex << _low;
-        }
-        else {
-            out << std::hex << _high << std::setw(16) << std::setfill('0')
-                << std::hex << _low;
-        }
+        out << std::setw(16) << std::setfill('0') << std::hex << _high
```
yurishkuro, line 46: this does not seem backwards compatible, what's the reason for removing `if (_high == 0) {`?

yurishkuro (commented):
Q: since you're looking into this, I just want too make sure there is a unit test that ensures that this library still CAN consume non-padded IDs from the incoming request headers

tobiasstadler:
> Q: since you're looking into this, I just want too make sure there is a unit test that ensures that this library still CAN consume non-padded IDs from the incoming request headers

TEST(SpanContext, testFromStream) tests both padded and unpadded trace ids

src/jaegertracing/TraceID.h
tobiasstadler: done

tobiasstadler: I though the trace id is now always 32 hex chars long. But I was wrong.

yurishkuro: No, all Jaeger SDKs still use 64bit by default, you have to explicitly enable 128bit.

yurishkuro (approved):
Thanks!

src/jaegertracing/TraceID.h
tobiasstadler: Yes, noticed that after your comment. I misread the prs for the other sdks.

PR #260, opened 2021-02-09
GitHub Actions should execute tests

tobiasstadler:
- Currently GitHub Actions only build the library but do not execute the tests

- Added a test step to the workflows
- Fixed the tests

### DIFF
--- a/src/jaegertracing/testutils/TracerUtil.h
+++ b/src/jaegertracing/testutils/TracerUtil.h
@@ -41,7 +41,7 @@ struct ResourceHandle {
 };
 
 std::shared_ptr<ResourceHandle> installGlobalTracer();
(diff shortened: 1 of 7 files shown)


src/jaegertracing/testutils/TracerUtil.h
```diff
@@ -41,6 +41,7 @@ struct ResourceHandle {
 };
 
 std::shared_ptr<ResourceHandle> installGlobalTracer();
+std::shared_ptr<ResourceHandle> installGlobalTracer(bool traceId128Bit);
```
yurishkuro, line 44: `installGlobalTracer(true)` at call site is very hard to read, I would rather call this explicitly `installGlobalTracer128Bit`

tobiasstadler: done

yurishkuro (approved):
Thanks!

PR #257, opened 2021-01-05
Bump yaml cpp to 0.6.3

tinchi:
## Which problem is this PR solving?
- While i was trying to build nginx-ingress with open-tracing and jaeger on aplin:3.11 and 3:12 i run into: 
```
Error: exit status 1
2020/12/15 11:13:14 [error] 85#85: Failed to load tracing library /usr/local/lib/libjaegertracing_plugin.so: Error relocating /usr/local/lib/libjaegertracing_plugin.so: _ZN4YAML6detail9node_data12empty_scalarB5cxx11E: symbol not found
nginx: [error] Failed to load tracing library /usr/local/lib/libjaegertracing_plugin.so: Error relocating /usr/local/lib/libjaegertracing_plugin.so: _ZN4YAML6detail9node_data12empty_scalarB5cxx11E: symbol not found
nginx: configuration file /tmp/nginx-cfg667822866 test failed
```
Which seems due to both of them shipping with yaml-cpp | 0.6.3
https://pkgs.alpinelinux.org/packages?name=yaml-cpp&branch=v3.11
https://pkgs.alpinelinux.org/packages?name=yaml-cpp&branch=v3.12

## Short description of the changes
- Bump hunter/archive to v0.23.289
- changed APPEND from yaml-cpp::yaml-cpp to yaml-cpp

PR #268, opened 2021-04-29
Update hunter

forestsword:

The problem @tinchi described in https://github.com/jaegertracing/jaeger-client-cpp/pull/257 

In addition to that while trying to upgrade this library and opentracing in Envoy, the older version of yaml-cpp would cause seg faults when the config was parsed.

Just updates hunter and boost to the latest versions.


forestsword:
Closing to clear (maybe) the maintainer approval after reverting changes to the workflows.

PR #275, opened 2021-07-11
add more ways to set tracer tags

### LINKED ISSUE #274: Add support for adding tracer tags via config yaml

Currently config's tags (and tracer's tags) support setting by `JAEGER_TAGS` env with `fromEnv` method. However in some use cases there's no easy way to pass environment variable to it, only code/config yaml can be altered. 

In our scenario we need to add a specific tag (process tag) to the span, but also can not set environment variable as we like. So we wish to have a way to set the tags in another way. 

We propose to add tags parsing logic from config yaml. That is, a config like this:

```yaml
disabled: false
reporter:
  logSpans: true
tags:
  tag1: 1337
  tag2: some value
sampler:
  type: const
  param: 1
```
will render the resulting span looks like this(converted by opentelemetry's jaeger receiver, so that's why we need it in process tags):

image

This only requires adding something like below in `Config.h`:
```cpp
    static std::vector<Tag> parseTags(const YAML::Node node) {
        if (!node.IsDefined() || !node.IsMap()) {
          return std::vector<Tag>();
        }
        std::vector<Tag> tags;
        for (auto it=node.begin(); it != node.end(); it++) {
          std::string key = it->first.as<std::string>();
          std::string value = it->second.as<std::string>();
          tags.emplace_back(key, value);
        }
        return tags;
    }
```

Better yet, also providing some method like `AddTag` or `SetTag` for config/tracer to directly modify the tags.

Supporting various value types other than string seems not easy (I'm not very experienced in cpp and the yaml-cpp lib :cry:). But just like `JAEGER_TAGS` which set all value as string, we consider this is ok. 

If this is reasonable to add we can prepare a PR. WDYT?

Frefreak:
Resolves #274 

- enable parse tags from yaml (all value is string)
- add `setTag` function to set tag directly

### DIFF
--- a/src/jaegertracing/Config.cpp
+++ b/src/jaegertracing/Config.cpp
@@ -107,9 +107,4 @@ std::vector<Tag> Config::parseTags(const YAML::Node& node) {
     return tags;
 }
 
-void Config::setTag(const std::string& key, Tag::ValueType&& value)
-{
-    _tags.push_back(Tag(key, value));
-}
-
 }  // namespace jaegertracing
--- a/src/jaegertracing/Config.cpp
+++ b/src/jaegertracing/Config.cpp
@@ -92,7 +92,22 @@ Config::parsePropagationFormat(std::string strPropagationFormat)
     return propagationFormat;
 }
 
-void Config::SetTag(const std::string& key, Tag::ValueType&& value)
+std::vector<Tag> Config::parseTags(const YAML::Node& node) {
+    if (!node.IsDefined() || !node.IsMap()) {
+        return std::vector<Tag>();
+    }
+    std::vector<Tag> tags;
+    for (auto it=node.begin(); it != node.end(); it++) {
+        if (it->second.IsScalar()) {
+          std::string key = it->first.as<std::string>();
+          std::string value = it->second.as<std::string>();
+          tags.emplace_back(key, value);
+        }
+    }
+    return tags;
+}
+
+void Config::setTag(const std::string& key, Tag::ValueType&& value)
 {
     _tags.push_back(Tag(key, value));
 }
--- a/src/jaegertracing/Config.cpp
+++ b/src/jaegertracing/Config.cpp
@@ -92,4 +92,9 @@ Config::parsePropagationFormat(std::string strPropagationFormat)
     return propagationFormat;
 }
 
+void Config::SetTag(const std::string& key, Tag::ValueType&& value)
+{
+    _tags.push_back(Tag(key, value));
+}
+
 }  // namespace jaegertracing
--- a/src/jaegertracing/Config.h
+++ b/src/jaegertracing/Config.h
@@ -128,9 +128,6 @@ class Config {
 
     const std::vector<Tag>& tags() const { return _tags; }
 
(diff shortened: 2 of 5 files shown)


src/jaegertracing/Config.h
```diff
@@ -126,6 +128,8 @@ class Config {
 
     const std::vector<Tag>& tags() const { return _tags; }
 
+    void SetTag(const std::string& key, Tag::ValueType&& value);
```
yurishkuro, line 131:
```suggestion
    // set a tracer-level tag
    void setTag(const std::string& key, Tag::ValueType&& value);
```

src/jaegertracing/Tag.h
```diff
@@ -18,6 +18,7 @@
 #define JAEGERTRACING_TAG_H
 
 #include "jaegertracing/Compilers.h"
+#include "jaegertracing/utils/YAML.h"
```
yurishkuro, line 21: > This is not a great dependency for this file, please keep yaml related code in config.

yurishkuro:
Windows build is failing on the new test

src/jaegertracing/Config.cpp
```diff
@@ -92,4 +92,24 @@ Config::parsePropagationFormat(std::string strPropagationFormat)
...
+        if (it->second.IsScalar()) {
+          std::string key = it->first.as<std::string>();
+          std::string value = it->second.as<std::string>();
+          tags.emplace_back(key, value);
+        }
+    }
+    return tags;
+}
+
+void Config::setTag(const std::string& key, Tag::ValueType&& value)
```
yurishkuro, line 110: I think the build is still failing (on Ubuntu).

Is this method even needed in this PR? If the goal is to allow tags configured via YAML, then it's already taken care of in parseTags.

If the method is needed as a programmatic way of setting tags, does the value need to be `Tag::ValueType&&`? The Span uses `opentracing::Value&` https://github.com/jaegertracing/jaeger-client-cpp/blob/4d264a42f54593fdea17ab3063e373e957d25cb3/src/jaegertracing/Span.h#L167-L168
Or we could just restrict to accepting string values only.

Frefreak: I think its a "nice to have", but for the main purpose of this PR that's totally fine to remove this for now. Will do the modification soon.

src/jaegertracing/Config.h
```diff
@@ -126,11 +128,15 @@ class Config {
 
     const std::vector<Tag>& tags() const { return _tags; }
 
+    // set a tracer-level tag
+    void setTag(const std::string& key, Tag::ValueType&& value);
```
yurishkuro, line 132: this should go away too?

Frefreak: yep, sorry :cry: Fixed.

yurishkuro:
I don't know why hunter is failing on ubuntu

Frefreak:
After some googling this seems to be somewhat related to the versions of gcc and gtest. In my local linux environment actually there would be another error about "maybe-uninitialized" if use gcc 11 so I was using clang to build and test and could not re-produce this.

PR #278, opened 2021-07-20
cmake: cleanup Findthrift cmake method

ideepika:
we do not use thrift_compiler, removed it and updated cmake to our own
version, with required target artificats only.

Signed-off-by: Deepika Upadhyay <dupadhya@redhat.com>

- 

-


ideepika:
@yurishkuro @mdouaihy

yurishkuro:
Not sure I follow the objective.

Also, most other Jaeger SDK repos are using the Docker image that we built that contains Thrift compiler, instead of relying on the compiler to be installed on the machine.

PR #280, opened 2021-08-07
Do not block report when flushing spans

### LINKED ISSUE #279: `RemoteReporter::report()` is blocked

Hi Jaeger developers!
I'm trying to issue a performance problem. 

The relevent code lies here: https://github.com/jaegertracing/jaeger-client-cpp/blob/master/src/jaegertracing/reporters/RemoteReporter.cpp#L85

The problem is that `RemoteReporter::report()` is blocked when `RemoteReporter::sweepQueue()` is flushing spans to local agents/http endpoints because the two functions share the same `_mutex`. In production environment, we usually wish tracing reports act as fast as possible and UDP/HTTP network transports are too heavy.

Unlock `_mutex` before flushing spans. See PR.

zhangnaifan:
Signed-off-by: Zhang Naifan <naifanzhang@qq.com>

- Resolves #279

- Unlock `_mutex` before flushing spans through UDP/HTTP network transports.

### DIFF
--- a/src/jaegertracing/reporters/RemoteReporter.cpp
+++ b/src/jaegertracing/reporters/RemoteReporter.cpp
@@ -95,9 +95,11 @@ void RemoteReporter::sweepQueue() noexcept
                 const auto span = _queue.front();
                 _queue.pop_front();
                 --_queueLength;
+                lock.unlock();
                 sendSpan(span);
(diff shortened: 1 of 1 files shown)


yurishkuro:
library is deprecated

